PR #36, opened 2020-11-26
Fix commands + registries

shedaniel:
I need some testing done, please don't merge until that is done


shedaniel:
Seems to work on fabric and forge in production.

PR #116, opened 2021-04-09
Annotated All EventJS subclasses and classes.

Hunter19823:
-Added KubeJSEvent annotation to docs.
           -Attempted to annotated every EventJS class with appropriate event id.
           -Added GameStagesIntegration.GAMESTAGE_ADDED and GameStagesIntegration.GAMESTAGE_REMOVED
           -Added todo to remove KubeJSEvents.COMMAND_REGISTRY , KubeJSEvents.BLOCK_DROPS .
           -Added KubeJSEvents.SERVER_DATAPACK_FIRST_PRIORITY and KubeJSEvents.SERVER_DATAPACK_LAST_PRIORITY
           -Added KubeJSEvents.BLOCK_LOOT_TABLES
           -Added Deprecated annotation to BlockDropsEventJS
           -Added Deprecated annotation to CommandRegistryEventJS
           -Added Deprecated annotation OldItemTooltipEventJS


common/src/main/java/dev/latvian/kubejs/script/data/DataPackEventJS.java
```diff
@@ -10,6 +12,14 @@
 /**
  * @author LatvianModder
  */
+@KubeJSEvent(
+		server = {
+				KubeJSEvents.SERVER_DATAPACK_LOW_PRIORITY,
+				KubeJSEvents.SERVER_DATAPACK_HIGH_PRIORITY,
+				"server.datapack.first",
```
MaxNeedsSnacks, line 19: Shouldn't these be changed to the event IDs you've added to KubeJSEvents?

common/src/main/java/dev/latvian/kubejs/server/TagEventJS.java
```diff
@@ -26,6 +27,10 @@
 /**
  * @author LatvianModder
  */
+@KubeJSEvent(
+		server = { "<ItemTypeName>.tags" },
+		client = { "<ItemTypeName>.tags" }
```
LatvianModder, line 38: Tag event doesnt get fired on client. Also I would rather list the 4 basic tags - item, block, entity_type, fluid

common/src/main/java/dev/latvian/kubejs/server/TagEventJS.java
```diff
@@ -26,6 +27,10 @@
 /**
  * @author LatvianModder
  */
+@KubeJSEvent(
+		server = { "<ItemTypeName>.tags" },
```
MaxNeedsSnacks, line 37: Not 100% on this, `ItemTypeName` isn't really too expressive

common/src/main/java/dev/latvian/kubejs/item/OldItemTooltipEventJS.java
```diff
@@ -10,6 +11,10 @@
 /**
  * @author LatvianModder
  */
+@Deprecated
+@KubeJSEvent(
+		client = { "client.item_tooltip" }
```
LatvianModder, line 16: I wouldn't include deprecated events

common/src/main/java/dev/latvian/kubejs/script/data/DataPackEventJS.java
LatvianModder, line 19: here too

LatvianModder (changes requested):

Hunter19823:
You know what, I accidentally reverted the wrong changes when I was fixing my local build.
I did have all those as events earlier, let me fix that really quickly.

common/src/main/java/dev/latvian/kubejs/script/data/DataPackEventJS.java
Hunter19823: Fixed in the latest commit.

common/src/main/java/dev/latvian/kubejs/server/TagEventJS.java
Hunter19823: Fixed in latest commit.

common/src/main/java/dev/latvian/kubejs/item/OldItemTooltipEventJS.java
Hunter19823: I removed the reference to the ID, however, that event is still being fired under that ID.
Unlike the rest of the deprecated events.

Hunter19823:
The last commit should fix everything you requested to change.
I do want to note that some events that called event.post(ID) instead of event.post(ScriptType,ID) were tricky to figure out whether they were on server/client as they often just used the world type.
For those, I just marked them as both client and server.

So you might want to take a quick look through any client-specific or server-specific events you know of to see if I accidentally marked them as both client and server inaccurately.

Hunter19823:
I just removed the KubeJSEvents.BLOCK_DROPS since that is no longer an event.
As well as I removed my old todo comments.

PR #124, opened 2021-04-30
add mode annotation

MelanX:
As mentioned on discord [here](https://discord.com/channels/303440391124942858/615627049637380144/837630095174139905), I added a "annotation" for modes.
That way, you can set a mode to a script. If the mode does not match, it will be ignored. If no mode is set, the script will load anyways.


LatvianModder:
Can you rename it to `modpack_mode` or something more descriptive? I feel like `mode` is too vague

MelanX:
I used "Packmode" because there was a mod in 1.12.2 which was named like this. Most people should already know it or it should be self-explaining enough I hope.

NillerMedDild:
Would you perhaps consider reading the mode from a `mode.json` found in the modpack root? That way modes can be hotswapped with the mod Config Swapper, and using a command like `/mode expert`

PR #126, opened 2021-05-02
Add a proper README

MaxNeedsSnacks:
I figured our README could do with a bit of a doover, so here goes!

This new readme contains tips and guidelines which are *mostly* aimed at mod/addon devs and contributors, like for example explaining how to set up KubeJS as a ForgeGradle / Loom dependency and how to integrate existing mods with it (recipe addons, class filters, plugins, etc.), as well as providing some **very** basic contribution guidelines, which I would like to expand upon later in a proper `CONTRIBUTING.md` file later down the line.

Overall, this new README is supposed to lower the barrier of entry to creating things for and with KubeJS just a little bit more, and I *hope* I've managed to achieve that somewhat ^^

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,4 +149,25 @@ A newer, more convenient way to add bindings, set class filters and even add **N
 
 ## Contributing to KubeJS
 
-### WIP
+### Getting Started
+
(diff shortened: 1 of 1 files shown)


MaxNeedsSnacks:
Based on several second-hand opinions this seems to be fine and a good addition, so I'll move this to last call for now, so make any further amendments **now** if you want to.

PR #130, opened 2021-05-12
Post filter tags

shedaniel:
#abc:
- a
- b
- c

#abcd:
- #abc
- d

Removing c from #abc will result in c being absent in #abc, but still present in #abcd.
This event should only be used for unfortunate situations where this is absolutely the only way left, this event ignores hierarchy, and is a pure filtering transformation.


common/src/main/java/dev/latvian/kubejs/server/FilterTagEventJS.java
```diff
@@ -0,0 +1,53 @@
...
+	private final Map<ResourceLocation, Tag<T>> tagMap;
+	private final Function<ResourceLocation, Optional<T>> idToValue;
+
+	public FilterTagEventJS(Map<ResourceLocation, Tag<T>> tagMap, Function<ResourceLocation, Optional<T>> idToValue) {
+		this.tagMap = tagMap;
+		this.idToValue = idToValue;
+	}
+
+	private Tag<T> wrapTag(Tag<T> tag, Predicate<T> filter) {
+		return Tag.fromSet(tag.getValues().stream().filter(filter).collect(Collectors.toSet()));
```
MaxNeedsSnacks, line 24: I feel like rather than wrapping the tag with each step and putting it back in the map, it'd be wiser to just join the predicates together and then apply them all at once for each tag *after* the event has been posted (see EventJS#afterPosted)

MaxNeedsSnacks (changes requested):

common/src/main/java/dev/latvian/kubejs/server/FilterTagEventJS.java
shedaniel: I think it is easier to not do that instead, I wanted the predicates based on tags to be using the already filtered tags.

PR #174, opened 2021-08-03
Implement LivingEntityDropsExperienceEventJS for forge

LLytho:
Using `LivingExperienceDropEvent` from forge to implement the event. 

Event is cancel able which will drop no experience for the player.
`getXp()` can return a different xp value if its modified before through other mods or events. 
`getVanillaXp()` will return the experience the mod will normally drop.


MaxNeedsSnacks:
Thanks for the contribution! Quick question: Would it be possible for you to implement this on the Architectury side (if it doesn't already exist, I'm not 100% sure about that) so that this event can be cross-platform from the get go? Otherwise I'd probably just do this myself later if the topic comes up :p

lythowastaken:
Do I have a deja vu or didn't we talk about this already? Maybe in discord :D I forgot here to look into the fabric side to implement it then. And what other kind of mini events exists to implement. I started a small list with some events and I also would like too see some dimension check (or teleport because it should include dimensions too). Could be useful for the kubejs people to disable dimension teleports based on conditions.

PR #240, opened 2021-12-12
Make PropertyObject public in BlockIDPredicate

LLytho:
As `BlockIDPredicate::getBlockProperties` is public and returns a list of `PropertyObject` it would be nice if it's public too to have access to it from the outside. For an example a plugin.


MaxNeedsSnacks:
Oh jeez rebasing this did NOT work... can you redo this PR for 1.18 :P

PR #320, opened 2022-04-21
Add ItemPropertiesEventJS

LLytho:
Add the ability to declare item properties which can be used in your item models.

Into `startup_scripts`:
```js
onEvent("item.model_properties", (event) => {
	/**
         * You can use every notation for item like `Ingredient.of()` or `Item.of()`, 
         * which makes it possible too also use multiple items at once.
         * 
	 * Register item property for diamonds with the name "kubejs:hidden_emerald".
	 * level and entity can be null!
	 */
	event.register("minecraft:diamond", "hidden_emerald", (stack, level, entity, id) => {
		// example which checks if the nbt contains "is_emerald"
		if(stack.nbt && stack.nbt.is_emerald) {
			return 1;
		}
		return 0;
	});

	/**
	 * Register item property for all items with the name "kubejs:hidden_paper".
	 * level and entity can be null!
	 */
	event.registerAll("hidden_paper", (stack, level, entity, id) => {
		// example which checks if the nbt contains "is_hidden"
		if(stack.nbt && stack.nbt.is_hidden) {
			return 1;
		}
		return 0;
	})
});
```

Example to use in your model, in our example `diamond.json`. 
You can learn more about models and properties here: https://minecraft.fandom.com/wiki/Model#Item_predicates
```json
{
	"parent": "minecraft:item/generated",
	"textures": {
		"layer0": "minecraft:item/diamond"
	},
	"overrides": [
		{
			"predicate": {
				"kubejs:hidden_emerald": 1
			},
			"model": "minecraft:item/emerald"
		},
		{
			"predicate": {
				"kubejs:hidden_paper": 1
			},
			"model": "minecraft:item/paper"
		}
	]
}
```

image

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/KubeJSEvents.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/KubeJSEvents.java
@@ -82,6 +82,7 @@ public interface KubeJSEvents {
 	String ITEM_FOOD_EATEN = "item.food_eaten";
 	String ITEM_TOOLTIP = "item.tooltip";
 	String ITEM_MODIFICATION = "item.modification";
+	String ITEM_PROPERTIES = "item.properties";
 
 	String SOUND_REGISTRY = "sound.registry";
 }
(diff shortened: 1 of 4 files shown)


MaxNeedsSnacks:
One thing upfront — Please rename this to ItemModelPropertiesEventJS to avoid confusion ^^

"Properties" is a bit ambiguous in this case because *item properties* are basic settings used during item setup (max stack size, durability, etc.)

MaxNeedsSnacks (approved):

PR #346, opened 2022-05-08
Public properties in BlockIdPredicate

LLytho:


MaxNeedsSnacks (approved):
I have a feeling changing property or value in the PropertyObject doesn't affect the actual state right? If so this should be good to merge!

PR #348, opened 2022-05-09
Extract type wrapper for BlockPos and Vec3

LLytho:
Extract the Vec3 and BlockPos typewrapper so they can be used outside if needed


MaxNeedsSnacks:
I'd rather move these to Rhino directly to be honest, @LatvianModder what's your take on this

PR #356, opened 2022-05-16
Add a bunch of useful methods to BlockContainerJS

ChiefArug:
Added methods:
 - getInventory(), returns the InventoryJS of the block from the top. Makes it beanable for simple inventories like chests
 - getDrops(PlayerJS, ItemStackJS), returns a ListJS of ItemStackJS, which contain the results of rolling the blocks loot table, taking into account the players effects and the held items enchantments/properties
 - getDrops(), returns a ListJS of ItemStackJS, which contain the results of rolling the blocks loot table but not taking into account any external factors, beanable
 - popItem(ItemStackJS), pops an item from the block, like when you break a block
 - popItemFromFace(ItemStackJS, Direction), pops an item from the face with more velocity away from that face (useful for when there is still a full block there, to avoid it shooting out in a random direction)

Feedback is appreciated *first time writing java code w/o tutorial*


ChiefArug:
Closing this because of the extra commits I somehow hauled along, probably because i branched off 1.18/main but am pring into 1.18/dev.
Will resubmit later.

PR #391, opened 2022-06-21
Add modifyPool, removeItem/Table and clearEntries/Conditions/Functions methods for use in loot table modification

ChiefArug:
Take two
This is a WIP (mainly because I haven't tested it), im putting it here so that people can ping me to remind me to finish it.

The only thing that is not in here that maybe could be is support for removing other types of entries. Im not sure what the best way to go about that. (ForEachEntry function maybe? Would require a method in the entry for removing it from the parent though)

Feedback is welcome and appreciated.


ChiefArug:
Because LootJS is now on both Forge and Fabric, and is so much more functional than this I am going to close this PR.

If anyone is interested in this I may consider it, but LootJS is just so much better.

PR #397, opened 2022-06-25
Added many more materials utilising many missing sound types

Xaidee:
Added almost all of vanilla's sound types to corresponding materials.

This will give the player a larger variety of sounds/materials to choose from, for example I needed amethyst block sounds for some compatibility blocks i need in a modpack im working on but it was practically impossible to do that without some real jankiness.


MaxNeedsSnacks:
See my comments regarding this PR [here](https://github.com/KubeJS-Mods/KubeJS/pull/398#issuecomment-1172331167), we currently don't know whether this will be merged for 1.19 or not

PR #398, opened 2022-06-25
Added many more materials utilising many missing sound types

Xaidee:
Added almost all of vanilla's sound types to corresponding materials.

This will give the player a larger variety of sounds/materials to choose from, for example I needed amethyst block sounds for some compatibility blocks i need in a modpack im working on but it was practically impossible to do that without some real jankiness.

This is a duplicate of #397 for 1.18 (without the soundtypes that are introduced in 1.19)


MaxNeedsSnacks:
(Oh right, squash merging exists)

PR #407, opened 2022-07-02
Fix removing tags from tags

LLytho:
Current problem is that you can't remove tags inside a tag like:
```
#minecraft:logs
- #minecraft:logs_that_burn
- #minecraft:crimson_stems
- #minecraft:warped_stems
```

```js
onEvent("item.tags", (e) => {
	e.remove("minecraft:logs", "#minecraft:logs_that_burn");
});
```

Using the code would not remove the tag. The PR will fix this.

Result:
image


MaxNeedsSnacks:
...Damnit :P

PR #420, opened 2022-07-18
Make the error for type() a bit less confusing

ChiefArug:
As it is atm some people think you include the `id: ` and `type: ` bits, and the error for doing it like that is vague (`Error loading KubeJS script: missing ) after argument list`)

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/BuilderBase.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/BuilderBase.java
@@ -45,7 +45,7 @@ public abstract class BuilderBase<T> implements Supplier<T> {
 	public abstract T createObject();
 
 	public final BuilderBase<T> type(String type) {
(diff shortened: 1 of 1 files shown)


common/src/main/java/dev/latvian/mods/kubejs/BuilderBase.java
```diff
@@ -45,7 +45,7 @@ public BuilderBase(ResourceLocation i) {
 	public abstract T createObject();
 
 	public final BuilderBase<T> type(String type) {
-		throw new RuntimeException("type(type: string) is no longer supported! Use event.create(id: string, type: string) now!");
+		throw new RuntimeException("type(type String) is no longer supported! Use event.create(id String, type String) now!");
```
MaxNeedsSnacks, line 48: If we're gonna go for java-esque syntax, we should go for **java-esque** syntax (i.e. `String type`)

ChiefArug: Makes sense, fixed!

PR #457, opened 2022-08-20
Added way to create item name dynamically

Prunoideae:
Adds a way to generate item name dynamically by NBT or something else. It's especially useful when users want to create item like `%s of %s`, or `The %s Sword of %s`, or whatever.

Example Code:
```js
onEvent("item.registry", event => {
    event.create("sussy_baka")
        .name(itemStack => {
            return Component.of(`Sussy baka - ${Math.random()}`)
        })
        .texture("minecraft:item/sugar")
})
```

Note that this function is called for every tick.

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/item/ItemBuilder.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/item/ItemBuilder.java
@@ -92,6 +92,9 @@ public abstract class ItemBuilder extends BuilderBase<Item> {
 	public transient FoodBuilder foodBuilder;
 	public transient Function<ItemStackJS, Color> barColor;
 	public transient ToIntFunction<ItemStackJS> barWidth;
+
+	public transient Function<ItemStackJS, Component> getName;
+
 	public transient Multimap<ResourceLocation, AttributeModifier> attributes;
 	public transient UseAnim anim;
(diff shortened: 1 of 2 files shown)


common/src/main/java/dev/latvian/mods/kubejs/item/ItemBuilder.java
```diff
@@ -92,6 +92,9 @@ public static ArmorMaterial ofArmorMaterial(Object o) {
 	public transient FoodBuilder foodBuilder;
 	public transient Function<ItemStackJS, Color> barColor;
 	public transient ToIntFunction<ItemStackJS> barWidth;
+
+	public transient Function<ItemStackJS, Component> getName;
```
ChloeDawn, line 96: Due to runtime type erasure, this implementation will not support implicit component wrapping (e.g passing a native object to represent component JSON). This must be a concrete type, which you can find other examples of at the tail of the builder class.

ChloeDawn, line 96: A better name for this field would be `nameGetter`, as it only "gets" the name when dereferenced with the invocation method.

ChloeDawn (changes requested):

ChloeDawn (approved):

PR #459, opened 2022-08-22
Add Cardinally orientable blocks and shuffle asset generation code

ChiefArug:
```js
onEvent('block.registry', event => {
	event.create('test_block', 'cardinal')
		.texture('side', 'minecraft:block/furnace_top')
		.texture('front', 'minecraft:block/iron_block')
})
```

This also shuffles asset generation around sot it is now in five different methods.
`generateAssetJsons` no longer directly generates the assets. Instead it passes that on to `generateBlockStateJson`, `generateBlockModelJsons` and `generateItemModelJson`. This makes it so that subclasses only need to override the necessary methods.
The fifth method is `generateMultipartBlockstateJson` in `MultipartShapedBlockBuilder`, which should be used for blocks such as fences and walls that use multipart blockstates.
All KubeJS block builders now use this system, meaning that ones that used to remove the checks for the user overriding a model/blockstate json should no longer do that.
This should also not break addons that still override `generateAssetJsons`.


ChiefArug:
I can also port this to 1.19 if wanted, although i know the 6.0 rewrite is coming so that may need to wait

ChiefArug:
I have just realized a flaw in this, the Block object returned does not extend BasicBlockJS, so behaviours like random ticking and waterlogging do not function properly.
The solution is probably going to be reimplementing the facing blockstate on an extension of BasicBlockJS, instead of HorizontalDirectionalBlock.

ChiefArug:
Another potential problem: the hitbox does not rotate when the state changes. Should be an easy enough fix

ChiefArug:
This should be ready to merge now.

PR #484, opened 2022-09-28
Implement Block Entities

MasterOfBob777:
Implements Block Entities, does not implement Screens, Custom Render Models, or much else besides Capabilities (Forge), and custom tickers. This PR does not contain these as they are in and of themselves complex systems.

### Example

```js
  event
    .create('example_block')
    .material('wood')
    .hardness(1.0)
    .displayName('Example Block')
    .blockEntity((builder) => {
      builder
      .ability("inventory", {
        type: "item", // can be any of fluid, energy, or item
        slots: [
          {
            id: "slot1",
            limit: 64 * 3 * 7,
            input: true, // todo 
            output: true, // todo
            filter: "minecraft:dirt" // todo
          }
        ]
      })
      // inventory-gui sets up the player's inventory for you
      // NOTE: Currently not supported, but dummy method has been put in place to be used at a later date
      .createScreen(`<inventory-gui>
        <div style="justify-items: center;">
          <slot attach="inventory#slot1" />
          <br />
          <button id="empty">Empty Dirt</button>
        </div>
      </inventory-gui>`)
      // NOTE: Currently not supported, see above
      .onContentLoaded((ev) => {
        ev.document.querySelector("#empty").addEventListener("click", () => {
          ev.blockEntity.getAbility("inventory").setCount("slot1", 0);
        })
      })
      .tick(20, (ev) => { // runs function every X ticks
        Utils.server.tell("hi")
      })
    })
})
```

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/BuiltinKubeJSPlugin.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/BuiltinKubeJSPlugin.java
@@ -1,5 +1,5 @@
-package dev.latvian.mods.kubejs;
 
+package dev.latvian.mods.kubejs;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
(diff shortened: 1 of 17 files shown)


MichaelHillcox:
I'd be really interested to so the performance metrics of a system like this... not sure if I agree with the html / xml screen system and mixing client & server code into one bundle... other than that, it's an interesting concept.

MaxNeedsSnacks:
Marking this as a draft until it's in a testable / mergeable state. I'd say this is an interesting poc that certainly warrants further experimenting, though I honestly disagree with the use of XML / markup text in general here, and think it'd probably be much better to just have a gui builder class instead (as well as some other design notes that I will share as they come up here)

Prunoideae:
Many of the code changes are requiring special handlers on ProbeJS side to get proper typing. It is still arguable for the capability part considering that I'm recently working on an addon which allows addition of any exported capability to almost anything (including itemstacks, entities and etc.).

BobVarioa:
I was thinking about this PR yesterday and like truth is I am not gonna ever finish it (it's not that useful anyways), so I'm gonna close this PR

PR #495, opened 2022-10-08
Allow adding custom objects to registries.

ChiefArug:
This opens up a lot of possibilities with registering things based on other mods classes, including things that do not have a default ('basic') type, such as Block Entities or Entities.

Example script, which adds a custom waystone block and item. They are fully functional.
```js
const WaystoneBlock = java('net.blay09.mods.waystones.block.WaystoneBlock')
const BProperties = java('net.minecraft.world.level.block.state.BlockBehaviour$Properties')
const Material = java('net.minecraft.world.level.material.Material')
const BlockItem = java('net.minecraft.world.item.BlockItem')
const IProperties = java('net.minecraft.world.item.Item$Properties')

let thingo

	thingo = event.custom('epic_waystone', new WaystoneBlock(BProperties.of(Material.AMETHYST)))
})
StartupEvents.registry('item', event => {
	event.custom('epic_waystone', new BlockItem(thingo.get(), new IProperties()))
})
```
Additionally, there is `event.fromBuilder` that takes a BuilderBase, for situations where you want to base it off an existing builder but not around some of the restrictions it normal has (Like a BlockItemBuilder)

It is expected that scripters provide their own assets and data jsons for custom registry objects, as trying to generate them generically is not possible.

Tested and works on both forge and fabric

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/BuilderBase.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/BuilderBase.java
@@ -135,4 +135,39 @@ public abstract class BuilderBase<T> implements Supplier<T> {
 	protected final RegistrySupplier<T> asRegistrySupplier() {
 		return Objects.requireNonNull(object, () -> "Object '%s' of registry '%s' hasn't been registered yet!".formatted(id, getRegistryType().registryKey.location()));
 	}
+
+	// Class that acts as a holder for a custom registry object, usually based off a class from another mod.
+	static class CustomBuilderObject<T> extends BuilderBase<T> {
+
+		private final T object;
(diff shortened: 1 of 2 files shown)


ChiefArug:
This also works when cherrypicked to 1.18. I can make another pr for that if desired.

ChiefArug:
Another example on 1.18.2 adding a new Tinkers Construct tool, a shovel.
The ridiculous script size is due to the number of jsons tinkers requires, however the item can still be created without those (it will not be functional as a tinkers tool though). The script generates all required jsons.
[tinkers_shovel.zip](https://github.com/KubeJS-Mods/KubeJS/files/9740513/tinkers_shovel.zip)

PR #498, opened 2022-10-09
Custom Leaves

### LINKED ISSUE #499: Add custom leaves block builder to the registry.
### Describe the feature

Please add a new custom leaves block builder; this would allow new leaves blocks to be created and use the vanilla leaf decay mechanics.

### Additional info

This is hypothetically possible with functions, but doing so would be a pain and not very performant.

cofl:
Adds a custom leaves block builder with sensible default values.

Example usage:

```js
onEvent('block.registry', event => {
	event.create('bamboo_leaves', 'leaves')
		.displayName('Bamboo Leaves')
		.textureAll('minecraft:block/azalea_leaves')
})
```

Closes #499.


cofl:
I have tested this and found it to work on both Forge and Fabric with the above sample. Please let me know if you would like PRs for 1.19 as well.

PR #512, opened 2022-10-16
Player Mount Issue

ripsterdog:
Anytime an entity mounts a player, its collision, suffocation detection, etc. properly moves on top of the player but visually its model doesn't move until it dismounts


ChiefArug:
Pull requests are not for issues, please make an issue here: https://github.com/KubeJS-Mods/KubeJS/issues/new?template=bug-template.yml instead

PR #552, opened 2022-12-02
feature: Follow Symlinks when traversing scripts

impleri:
### Description

Adjust KubeJS.loadScripts to follow symlinks when traversing scripts.

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/KubeJS.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/KubeJS.java
@@ -33,6 +33,8 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.FileSystemLoopException;
+import java.nio.file.FileVisitOption;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
@@ -155,7 +157,7 @@ public class KubeJS {
 		final var pathPrefix = path;
 
 		try {
-			for (var file : Files.walk(dir, 10).filter(Files::isRegularFile).toList()) {
+			for (var file : Files.walk(dir, 10, FileVisitOption.FOLLOW_LINKS).filter(Files::isRegularFile).toList()) {
 				var fileName = dir.relativize(file).toString().replace(File.separatorChar, '/');
 
 				if (fileName.endsWith(".js")) {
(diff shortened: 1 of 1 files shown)


MaxNeedsSnacks:
Looks fine to me, I'm assuming the main use case for this would be to have cross-instance script folders for testing or something along those lines?

squoshi:
> Looks fine to me, I'm assuming the main use case for this would be to have cross-instance script folders for testing or something along those lines?

Ideally dev servers or multiple instances. I tried making a server with it but it didn't work.

MaxNeedsSnacks:
Hmm, can you give a little bit more detail about what the issue is here @squoshi? I'd like to see this merged with the other changes in the next release if possible

impleri:
I linked a folder within `server_scripts`. I'm not sure if `kubejs` itself can be symlinked.

MaxNeedsSnacks:
This PR shouldn't need to worry about that because it only addresses the script managers. So long as those work fine it's all good :P

impleri:
It's also working for me if `server_scripits` is a symlink. So should work for anything at or under the script manager root directory.

squoshi:
> Hmm, can you give a little bit more detail about what the issue is here @squoshi? I'd like to see this merged with the other changes in the next release if possible

KubeJS scripts didn't load. I don't know if it's a me problem or a KubeJS problem.
Checkmarks mean they work, X's mean they don't.
image

MaxNeedsSnacks:
Again, symlinking the kubejs folder is not the concern here, we’re talking about the folders inside the kubejs folder for now

MaxNeedsSnacks:
Looking for an update from @squoshi here, does the intended usecase (symlinking *subfolders* of ./kubejs/) work here?

MaxNeedsSnacks:
Mostly waiting on a response from @squoshi, also, in theory it'd be more practical if the "kubejs" folder itself was also symlinkable, wouldn't it?

impleri:
I just tested this and it indeed works for that effect.

ll-output

ll2-output

ll3-output

kube-log

Perhaps the issue is that a symlink is not the same as a Windows `.lnk` file?

MaxNeedsSnacks:
Ah right, I was assuming they were using directory junctions (the NTFS version of directory symlinks)

MaxNeedsSnacks:
Thanks for the contribution! We're currently still preparing for a new (possibly bigger) release of KubeJS so this change won't *immediately* be released, but it's now on the dev branch (and if applicable and I remember to do so, I'll be backporting it to 1.18 as well)

PR #598, opened 2023-03-22
Updated Antimatter API url 

Rabbitminers:
### Description 

Location of integration seems to have moved to common so the url doesn't work


LatvianModder:
You made PR to main branch, but it's fine, I'll just update it in dev branch, thanks

PR #608, opened 2023-04-15
Deprecate fullNBT methods and replace with just nbt

ChiefArug:
### Description 
This fixes a discrepancy between various things methods for getting the NBT data attached to that object, where entities always had getFullNBT instead of getNbt, setFullNBT instead of setNbt and mergeFullNBT instead of mergeNbt like ItemStackKJS has.

It does this by deprecating the old methods and adding an error to anyone using those, hopefully encouraging them to switch before they are removed (probably in 1.20)

The loss of full NBT capitalization is to prevent odd beaning (getNBT beans to nBT).

#### Other details 
This discrepancy comes from back when entities getNbt method return KubeJS' persistent data, which no one ever really used. Those methods were moved to the full name of getRawPersistentData and setRawPersistentData, leaving the simple getNbt and co unused.


ladylexxie:
Big big hype!

PR #609, opened 2023-04-15
Use placed block's state instead of default state in BlockStateModifyPlacementCallbackJS

### LINKED ISSUE #600: Use placed block's state instead of default state in BlockStateModifyPlacementCallbackJS
### Describe the feature

In `BlockStateModifyPlacementCallbackJS`, it passes in the block's default state in the `super` call. But shouldn't the block state of the placed block in the world be passed into the super?

I propose to change the super call in `BlockStateModifyPlacementCallbackJS` from
```java
public BlockStateModifyPlacementCallbackJS(BlockPlaceContext context, Block block) {
  super(block.defaultBlockState());
  this.context = context;
  this.minecraftBlock = block;
  this.block = new BlockContainerJS(context.getLevel(), context.getClickedPos());
}
```
to be
```java
public BlockStateModifyPlacementCallbackJS(BlockPlaceContext context, Block block) {
  super(blockPlaceContext.getLevel().getBlockState(blockPlaceContext.getClickedPos()));
  this.context = context;
  this.minecraftBlock = block;
  this.block = new BlockContainerJS(context.getLevel(), context.getClickedPos());
}
```

### Additional info

_No response_

ChiefArug:
### Description 
Fixes #600

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/block/BlockStateModifyPlacementCallbackJS.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/block/BlockStateModifyPlacementCallbackJS.java
@@ -23,7 +23,7 @@ public class BlockStateModifyPlacementCallbackJS extends BlockStateModifyCallbac
 	public BlockContainerJS block;
 
 	public BlockStateModifyPlacementCallbackJS(BlockPlaceContext context, Block block) {
-		super(block.defaultBlockState());
+		super(context.getLevel().getBlockState(context.getClickedPos()));
 		this.context = context;
 		this.minecraftBlock = block;
 		this.block = new BlockContainerJS(context.getLevel(), context.getClickedPos());
--- a/common/src/main/java/dev/latvian/mods/kubejs/block/BlockStateModifyPlacementCallbackJS.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/block/BlockStateModifyPlacementCallbackJS.java
@@ -10,6 +10,7 @@ import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.context.BlockPlaceContext;
 import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.Block;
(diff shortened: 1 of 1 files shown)


common/src/main/java/dev/latvian/mods/kubejs/block/BlockStateModifyPlacementCallbackJS.java
```diff
@@ -23,7 +23,7 @@ public class BlockStateModifyPlacementCallbackJS extends BlockStateModifyCallbac
 	public BlockContainerJS block;
 
 	public BlockStateModifyPlacementCallbackJS(BlockPlaceContext context, Block block) {
-		super(block.defaultBlockState());
+		super(context.getLevel().getBlockState(context.getClickedPos()));
```
MaxNeedsSnacks, line 26: Is it guaranteed that the block already exists at that position at that time?

ChiefArug: Ah, it is not. It is the previous block that was there

ChiefArug: I guess check if it is the previous block that was there, otherwise default state. 
@Spyeedy what was your intention with this suggested change? The only place this seems to be used is for custom placementStates for custom blocks, where the block you are placing is never the old block.

Spyeedy: Yes, my intention was to modify the already placed block's state, similar to how snow layer block works.

ChiefArug:
It now only uses the current state if it is the same as the block being placed, otherwise it uses the default state.
Logic for checking thatis in a separate method due to super constructor calls needing to be first, and not wanting a behemoth ternary operator.
There is also a method for checking if the block is replacing itself, so people can check.

PR #627, opened 2023-05-28
KubeJS 6.1 - The whole thing

MaxNeedsSnacks:
This pull request mostly exists for me to keep track of all the individual changes contained in KubeJS 6.1, as well as to perhaps get in some feedback from other developers about upcoming changes as well.


fabric/src/main/java/dev/latvian/mods/kubejs/fabric/KubeJSFabric.java
```diff
@@ -23,7 +25,20 @@ public void onInitialize() {
...
+
+				if (added > 0) {
+					deferredRegister.register();
+				}
+			}
+		}
+
+		for (var info : RegistryInfo.AFTER_VANILLA) {
 			var deferredRegister = DeferredRegister.create(KubeJS.MOD_ID, UtilsJS.cast(info.key));
 			int added = info.registerObjects(deferredRegister::register);
```
MaxNeedsSnacks: Any reason in particular we're using DeferredRegister here? On Fabric, DeferredRegister just registers directly to the registry iirc, so we could just replace it with the appropriate call immediately

PR #639, opened 2023-06-22
Add a new item type, "shears".

AnAwesomGuy:
this is a "continuation" of #637 bcuz im an idiot with github and didnt know that renaming a branch would "remove" it.
example:
```js
onEvent('item.registry', event => {
	event.create('test_shears', 'shears').speedBaseline(1.2).maxDamage(98)
})
```


AnAwesomGuy:
apparently lat wants me to port this to 1.19 so im closing it

PR #652, opened 2023-07-18
Add minRolls and maxRolls to OutputItem

BobVarioa:
These come up in addons quite a bit, and they shouldn't change anything output wise. Just for new mods, it lets them do this in a prettier way.

Before
```js
event.mod.recipe("input_item", [
  {
    item: Item.of("dirt").withChance(0.5),
    minRolls: 1,
    maxRolls: 5
  }
])
```

After:
```js
event.mod.recipe("input_item", [
  Item.of("dirt")
    .withChance(0.5)
    .minRolls(1)
    .maxRolls(5)
])
```

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/item/OutputItem.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/item/OutputItem.java
@@ -9,36 +9,48 @@ import net.minecraft.nbt.CompoundTag;
 import net.minecraft.world.item.ItemStack;
 
 public class OutputItem implements OutputReplacement {
-	public static final OutputItem EMPTY = new OutputItem(ItemStack.EMPTY, Double.NaN);
+	public static final OutputItem EMPTY = new OutputItem(ItemStack.EMPTY, Double.NaN, 0, 0);
 
-	public static OutputItem of(ItemStack item, double chance) {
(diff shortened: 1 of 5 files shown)


common/src/main/java/dev/latvian/mods/kubejs/item/OutputItem.java
```diff
@@ -9,36 +9,48 @@
...
 	public final double chance;
+	public final int minRolls;
+	public final int maxRolls;
 
-	protected OutputItem(ItemStack item, double chance) {
+	protected OutputItem(ItemStack item, double chance, int minRolls, int maxRolls) {
 		this.item = item;
 		this.chance = chance;
+		this.minRolls = minRolls;
+		this.maxRolls = maxRolls;
```
MaxNeedsSnacks, line 37: If this is introduced, we will need to add a constructor with the old (ItemStack, double) signature to avoid API breakage.

MaxNeedsSnacks (changes requested):

common/src/main/java/dev/latvian/mods/kubejs/item/OutputItem.java
BobVarioa: Or rather, added the old signature back

PR #672, opened 2023-07-25
More Typings

ChiefArug:
### Description 
Moar typings 🎉 

#### Other details 
In doing this I have noticed a lot of seemingly redundant code, particularly around BlockWrapper and BlockPredicate, but have not touched that for fear of breaking addons. I haven't bothered documenting it though.

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/bindings/ItemWrapper.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/bindings/ItemWrapper.java
@@ -3,6 +3,7 @@ package dev.latvian.mods.kubejs.bindings;
 import dev.latvian.mods.kubejs.item.ItemStackJS;
 import dev.latvian.mods.kubejs.level.FireworksJS;
 import dev.latvian.mods.kubejs.registry.KubeJSRegistries;
+import dev.latvian.mods.kubejs.typings.Info;
 import net.minecraft.core.NonNullList;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.resources.ResourceLocation;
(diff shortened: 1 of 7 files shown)


common/src/main/java/dev/latvian/mods/kubejs/bindings/ItemWrapper.java
```diff
@@ -13,62 +14,79 @@
...
 	static boolean exists(ResourceLocation id) {
 		return KubeJSRegistries.items().contains(id);
 	}
 
+	@Info("""
+		Checks if the passed in object is an ItemStack.
+		Note that this does not mean it will not function as an ItemStack if passed to something that requests one.
+		""")
 	static boolean isItem(@Nullable Object o) {
 		return o instanceof ItemStackJS;
```
MaxNeedsSnacks, line 91: just noticed this error while going through to review this, this should be instanceof ItemStack, shouldn't it?

ChiefArug: ItemStackMixin does implement ItemStackJS, but I'm not sure if that gets added to ItemStack as part of the mixin.

Prunoideae:
It would be better if documents were in markdown format.

common/src/main/java/dev/latvian/mods/kubejs/bindings/ItemWrapper.java
MaxNeedsSnacks: it implemented ItemStack__KJS__ not ItemStackJS; ItemStackJS is just the leftovers of the item stack wrapper iirc

ChiefArug: Oh whoops, yeah. Do you want me to fix that in this PR?

ChiefArug:
Do the docs need formatting at all?

Prunoideae:
VSCode renders docs as markdown by default, though.

ChiefArug:
@LatvianModder are you all good with having @Info use markdown? Or should it be left as plain text.
Also, should I include examples for the `of` methods (ie ItemWrapper#of, TextWrapper#of(

ChiefArug:
I'm marking this as ready for merging now, but may continue to push commits if I feel like documenting anything else.

PR #674, opened 2023-07-28
Throw an informative error about null item stacks as recipe results

ChiefArug:
### Description 
ItemStacks should never be null, but some people are not aware of that. If someone does return null from `getResultItem` in a recipe then the [resulting error](https://gnomebot.dev/paste/1134361831868219452#L19) says nothing about the recipe that causes it. This fixes that by throwing earlier and including the recipe id.

#### Example Script 
```js
//requires: butchercraft
// specifically version 2.2.1 or lower, they have fixed it in newer versions
SeverEvents.recipes(event => {
  event.remove({output: 'minecraft:stick'}) // any filter that includes butchercrafts meat_hook recipes causes it
})
```

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/recipe/JsonRecipeJS.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/recipe/JsonRecipeJS.java
@@ -35,6 +35,8 @@ public class JsonRecipeJS extends RecipeJS {
 	public boolean hasOutput(ReplacementMatch match) {
 		if (CommonProperties.get().matchJsonRecipes && match instanceof ItemMatch m && getOriginalRecipe() != null) {
 			var r = getOriginalRecipe().getResultItem();
+			//noinspection ConstantValue
+			if (r == null) throw new NullPointerException("ItemStack should never be null, but recipe " + getId() + " returned null as the output!");
 			return r != ItemStack.EMPTY && m.contains(r);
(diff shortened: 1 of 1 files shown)


common/src/main/java/dev/latvian/mods/kubejs/recipe/JsonRecipeJS.java
```diff
@@ -35,6 +35,8 @@ public boolean replaceInput(ReplacementMatch match, InputReplacement with) {
 	public boolean hasOutput(ReplacementMatch match) {
 		if (CommonProperties.get().matchJsonRecipes && match instanceof ItemMatch m && getOriginalRecipe() != null) {
 			var r = getOriginalRecipe().getResultItem();
+			//noinspection ConstantValue
+			if (r == null) throw new NullPointerException("ItemStack should never be null, but recipe " + getId() + " returned null as the output!");
```
MaxNeedsSnacks, line 39: How does this error out during a recipe event? Is it a hard crash or does it continue as usual? Since this is not a user error, we would probably want to make sure it's the latter!

ChiefArug: It is not a hard crash. From what I can tell it stops the current loop over all recipes (ie that one remove). Any recipes that were iterated before this are still affected.

PR #675, opened 2023-07-28
More block callbacks

ChiefArug:
### Description 
I looked through the remaining overrides for BasicBlockJS and these are the ones that were left and made sense to me.

Also includes some slightly related changes, like a default value override for requiresTool() and making safeCallback not concat the string till the error is actually thrown (this was done because bounce is relatively hot code and i felt like mirco-optimising. and it does save a whole registry lookup)

#### Example Script 
```js
    event.create('double_slime')
        .bounciness(2)
        .exploded(ctx => {
            ctx.block.set('slime_block')
        })
        .stepOn(ctx => {
            if (ctx.entity.player)
                ctx.entity.playSound("block.slime_block.step", 0.8, 2.0)
        })

    event.create('flinger')
        .bounce(ctx => {
            ctx.bounce(2 * JavaMath.random())
        })
        .fallOn(ctx => {
            if (ctx.suppressingBounce) ctx.applyFallDamage()
        })

    event.create('slime_block') // bouncier slime
        .model('block/slime_block').defaultTranslucent().material("slime")
        .bounce(ctx => ctx.bounce(3))


    event.create('minecraft:coal_block') //registry replace coal so it can compact into diamonds
        .hardness(5).resistance(6).material("stone").requiresTool()
        .fallOn(ctx => {
            if (ctx.fallHeight > 20) {
                ctx.block.set('diamond_block')
            } else if (!ctx.level.clientSide) {
                ctx.entity.tell("There wasn't enough force! Try falling from higher")
                ctx.noBounce()
            }
        })
})
```

#### Other details 
As part of this I moved the block callbacks to their own package (it was getting crowded in the block package), so any addons using those classes will break. I have not checked if there are any.

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/block/BlockBuilder.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/block/BlockBuilder.java
@@ -680,8 +680,12 @@ public abstract class BlockBuilder extends BuilderBase<Block> {
 		return bounce(ctx -> ctx.bounce(-bounciness));
 	}
 
-	//TODO: is this too laggy? 1 tps isnt much...
-	@Info("Set how this block bounces/moves entities that land on top of this. Do not use this to modify the block, use fallOn instead!")
+	@Info("""
+		Set how this block bounces/moves entities that land on top of this. Do not use this to modify the block, use fallOn instead!
+		Use ctx.bounce(height) or ctx.setVelocity(x, y, z) to change the entities velocity.
+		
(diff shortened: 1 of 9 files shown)


ChiefArug:
This is a draft because I am getting very occasional game freezes from blocks with custom bounce behaviour, then it unfreezes a while (5+ mins) later and im at -3 billion y. I may end up removing the bounce functionality from this pr if I cannot fix that.

Prunoideae (approved):

Issue #474, opened 2022-09-07
Support for Multitool creation


4AC11:
### Describe the feature

When creating a tool using KubeJS, it is currently only possible to assign it one tool type out of the vanilla options, and it would be a helpful feature to be able to allow an item to register as multiple tool types, as many paxel/AIOT mods do.

### Additional info

In my personal example, I was hoping to create the Pickaxe Axe from terraria, before realizing that I could only choose one option for its tool type. I think it would be very helpful for anyone actually trying to create unique tools using KubeJS.

4AC11:
any progress on this?

AnAwesomGuy:
ill take a try at this

PR #681, opened 2023-08-01
Implement #474 and add a 'multitool

AnAwesomGuy:
#### Example Script 
```js
// priority: 0
StartupEvents.registry('item', event => {
    event.create('axe_shovel', 'multitool').tool('axe').tool('shovel').maxDamage(172)
    event.create('pickaxe_axe', 'multitool').tier('diamond').tool('axe').maxDamage(140).tool('pickaxe')
    event.create('iron_paxel', 'multitool').tool('pickaxe').tool('axe').tool('shovel').tool('hoe').maxDamage(64)
})
```

#### Other details \

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/item/custom/MultitoolItemJS.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/item/custom/MultitoolItemJS.java
@@ -1,0 +1,179 @@
+package dev.latvian.mods.kubejs.item.custom;
+
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+import dev.latvian.mods.kubejs.item.MutableToolTier;
+import dev.latvian.mods.kubejs.registry.KubeJSRegistries;
+import dev.latvian.mods.kubejs.typings.Info;
+import dev.latvian.mods.kubejs.typings.Param;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.tags.BlockTags;
+import net.minecraft.tags.TagKey;
+import net.minecraft.world.InteractionResult;
+import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.EquipmentSlot;
+import net.minecraft.world.entity.ai.attributes.Attribute;
+import net.minecraft.world.entity.ai.attributes.AttributeModifier;
+import net.minecraft.world.item.DiggerItem;
+import net.minecraft.world.item.Item;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.Items;
+import net.minecraft.world.item.context.UseOnContext;
+import net.minecraft.world.level.block.Block;
+import net.minecraft.world.level.block.state.BlockState;
+import org.jetbrains.annotations.Nullable;
+
+import java.math.BigDecimal;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+import static java.math.RoundingMode.CEILING;
+
+public class MultitoolItemJS extends DiggerItem {
+    public static class Builder extends HandheldItemBuilder {
+        protected boolean isAxe = false, isHoe = false, isPickaxe = false, isShovel = false;
+        protected final List<Float> speedValues = new ArrayList<>(5);
+        protected final List<Float> attackValues = new ArrayList<>(5);
+        protected final Set<TagKey<Block>> mineableTags = new HashSet<>();
+        public Builder(ResourceLocation i) {
+            super(i, 0, 0);
+        }
+        @Info(value = """
+            Adds a tool to the multi-tool.
+            
+            Valid tool types include: 'axe', 'hoe', 'pickaxe', and 'shovel'.
+            """, params =
+        @Param(name = "tool", value = "The name of the tool to add to the multi-tool. Will error if it is an unknown tool type."))
+        public Builder tool(String tool) {
+            // maybe swords and shears (after #673 is merged)?
+            return switch (tool) {
+                default -> throw new IllegalArgumentException("Unknown tool type '" + tool +
+                        "', valid tool types are: 'axe', 'hoe', 'pickaxe', and 'shovel'!");
+                case "axe" -> {
+                    isAxe = true;
+                    yield addValues(6, -3.1f, BlockTags.MINEABLE_WITH_AXE);
+                }
+                case "hoe" -> {
+                    isHoe = true;
+                    yield addValues(-2, -1, BlockTags.MINEABLE_WITH_HOE);
+                }
+                case "pickaxe" -> {
+                    isPickaxe = true;
+                    yield addValues(1, -2.8f, BlockTags.MINEABLE_WITH_PICKAXE);
+                }
+                case "shovel" -> {
+                    isShovel = true;
+                    yield addValues(1.5f, -3, BlockTags.MINEABLE_WITH_SHOVEL);
+                }
+            };
+        }
+        protected Builder addValues(float attack, float speed, @Nullable TagKey<Block> tag) {
+            attackValues.add(attack);
+            speedValues.add(speed);
+            if (tag != null)
+                mineableTags.add(tag);
+            return this;
+        }
+        @Override
+        public Builder attackDamageBaseline(float f) {
+            attackValues.clear();
+            attackValues.add(f);
+            return this;
+        }
+        @Override
+        public Builder speedBaseline(float f) {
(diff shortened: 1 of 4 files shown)


common/src/main/java/dev/latvian/mods/kubejs/item/custom/MultitoolItemJS.java
```diff
@@ -0,0 +1,179 @@
...
+                case "pickaxe" -> {
+                    isPickaxe = true;
+                    yield addValues(1, -2.8f, BlockTags.MINEABLE_WITH_PICKAXE);
+                }
+                case "shovel" -> {
+                    isShovel = true;
+                    yield addValues(1.5f, -3, BlockTags.MINEABLE_WITH_SHOVEL);
+                }
+            };
+        }
```
ChiefArug, line 72: Should this be an Enum?

common/src/main/java/dev/latvian/mods/kubejs/item/custom/MultitoolItemJS.java
```diff
@@ -0,0 +1,179 @@
...
+    @Override
+    public Multimap<Attribute, AttributeModifier> getDefaultAttributeModifiers(EquipmentSlot equipmentSlot) {
+        if (!modified) {
+            modified = true;
+            attributes.forEach((r, m) -> defaultModifiers.put(KubeJSRegistries.attributes().get(r), m));
+        }
+        return super.getDefaultAttributeModifiers(equipmentSlot);
+    }
+    @Override
+    // right-clicking on dirt will till, unless sneaking, then it will turn it into a path
```
ChiefArug, line 155: Is this consistent with other mods with multitools?

common/src/main/java/dev/latvian/mods/kubejs/item/custom/MultitoolItemJS.java
```diff
@@ -0,0 +1,179 @@
...
+            (isHoe && Items.IRON_HOE.useOn(ctx) != InteractionResult.PASS) ||
+            (isPickaxe && Items.IRON_HOE.useOn(ctx) != InteractionResult.PASS) ?
+            InteractionResult.sidedSuccess(ctx.getLevel().isClientSide) : InteractionResult.PASS;
+    }
+    @Override
+    public boolean isCorrectToolForDrops(BlockState state) {
+        final int i = tier.getLevel();
+        if ((i < 3 && state.is(BlockTags.NEEDS_DIAMOND_TOOL)) ||
+            (i < 2 && state.is(BlockTags.NEEDS_IRON_TOOL)) ||
+            (i < 1 && state.is(BlockTags.NEEDS_STONE_TOOL))) return false;
```
ChiefArug, line 170: Mods can add more tiers, even Forge by default adds gold and netherite as ones.
This should really handle those as well. not be hardcoded to vanilla.

fabric/src/main/java/dev/latvian/mods/kubejs/fabric/BuiltinKubeJSFabricPlugin.java
```diff
@@ -1,10 +1,18 @@
...
+import dev.latvian.mods.kubejs.registry.RegistryInfo;
 import dev.latvian.mods.kubejs.script.ScriptType;
 import dev.latvian.mods.kubejs.util.ClassFilter;
 
 public class BuiltinKubeJSFabricPlugin extends BuiltinKubeJSPlugin {
+	@Override
+	public void init() {
+		super.init();
+
+		RegistryInfo.ITEM.addType("multitool", MultitoolItemJS.Builder.class, MultitoolItemJS.Builder::new);
```
ChiefArug, line 14: Why only Fabric?

ChiefArug:
Another thing to consider is Forges ToolAction system. It would require a subclass for Forge tho.

fabric/src/main/java/dev/latvian/mods/kubejs/fabric/BuiltinKubeJSFabricPlugin.java
AnAwesomGuy: i also have [a class for forge](https://github.com/KubeJS-Mods/KubeJS/pull/681/commits/3f88795d68c501b3aa29e1509133c67c1902d9c4#diff-e5e3ace2152650752dcbf461f578a3cb672d62571d477b479a5bf7a03568174a), since it needs the toolaction things

common/src/main/java/dev/latvian/mods/kubejs/item/custom/MultitoolItemJS.java
AnAwesomGuy: im not sure, but idk how i would do it otherwise
(i couldn't find any paxel mods that can till and hoe at the same time)

common/src/main/java/dev/latvian/mods/kubejs/item/custom/MultitoolItemJS.java
AnAwesomGuy: if an enum was the arg, would rhino convert a string into the enum value?
bcuz if not, i dont really understand why this would need to be an enum or how it would work as an enum

ChiefArug: yes, rhino autowraps strings to enums. You could also then use an EnumSet instead of seperate booleans for each tool type, and have the addValues function be function on the enum, which I think would completely remove the need for a switch statement.

fabric/src/main/java/dev/latvian/mods/kubejs/fabric/BuiltinKubeJSFabricPlugin.java
ChiefArug: ah, I hadn't seen that

common/src/main/java/dev/latvian/mods/kubejs/item/custom/MultitoolItemJS.java
MaxNeedsSnacks: To be fair, this is a larger issue with KubeJS as a whole, since harvest levels are... pretty terribly done across Vanilla, Fabric and Forge. I'll need to find an abstraction for it, maybe even add it to Architectury

AnAwesomGuy: ive done a fix on forge, but have no idea how to do this on fabric

ChiefArug: Fabric has number based tiers for everything thats not vanilla iirc, so you can just append the passed in number to a string to get the tag if its not a default one. Their wiki probably has a page on it.

AnAwesomGuy:
i think its good now!

common/src/main/java/dev/latvian/mods/kubejs/item/custom/MultitoolItemJS.java
```diff
@@ -0,0 +1,198 @@
...
+        public Builder speedBaseline(float f) {
+            speedValues.clear();
+            speedValues.add(f);
+            return this;
+        }
+
+        protected void setValues() {
+            // TOD0: find a way to make this better / more efficient
+            // i hate floating-point jank
+
```
ChiefArug, line 76: You could multiply by 10, round, cast to int, do the math with ints, then divide by 10f back to float. You will lose a tiny bit of precision, but you are already rounding later on.
It would still be messy though, not sure theres much you could do about that.

common/src/main/java/dev/latvian/mods/kubejs/item/custom/MultitoolItemJS.java
```diff
@@ -0,0 +1,198 @@
...
+        protected final EnumSet<ToolTypes> toolTypes = EnumSet.noneOf(ToolTypes.class);
+
+        public Builder(ResourceLocation i) {
+            super(i, 0, 0);
+        }
+
+        @Info(value = """
+            Adds a tool to the multi-tool.
+            
+            Valid tool types include: 'axe', 'hoe', 'pickaxe', and 'shovel'.
```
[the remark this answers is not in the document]
AnAwesomGuy: it works as lowercase (i always test things)

common/src/main/java/dev/latvian/mods/kubejs/item/custom/MultitoolItemJS.java
AnAwesomGuy: im having a bit of trouble with this

LatvianModder:
Reopen if still relevant to current mc version

PR #694, opened 2023-08-27
Add missing getHand() to block left clicking event

ladylexxie:


ChiefArug:
erm, you can't left click with anything other than your main hand...

liopyu:
Screenshot 2023-08-27 202443
She posted because the blockleftclick event always fires twice in which we thought it was because it fired for each hand. When we tried solving it by trying to detect the hand "used" to left click she found the InteractionHand was never set. We later found out the real reason was because it fires once for the left click and once when you release left click for some reason.

PR #696, opened 2023-09-04
Fix name resolving of EnumComponent

rlnt:
The current implementation of the EnumComponent using `String.valueOf` doesn't correctly resolve the name of the current enum value.


rlnt:
superseded by https://github.com/KubeJS-Mods/KubeJS/commit/99dad816adf3bb09611548933efa0de5f7a7d205

PR #715, opened 2023-10-01
Fix serverOnly mode not working due to registering the new menu type

ChiefArug:
### Description 
Fixes [KubeJS 2001.6.3-build50 | serverOnly=true | Still requires me to have kubejs on client (BC1A0B2D)](https://kubejs.com/support/tickets/BC1A0B2D), caused by KubeJS' menutype always registering, even if serverOnly is enabled.

#### Other details

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/gui/KubeJSMenu.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/gui/KubeJSMenu.java
@@ -16,7 +16,7 @@ import java.util.function.Supplier;
 
 public class KubeJSMenu extends AbstractContainerMenu {
 	public static final DeferredRegister<MenuType<?>> MENU_TYPES = DeferredRegister.create("kubejs", Registries.MENU);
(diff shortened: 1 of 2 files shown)


common/src/main/java/dev/latvian/mods/kubejs/gui/KubeJSMenu.java
```diff
@@ -15,10 +16,12 @@
 
 public class KubeJSMenu extends AbstractContainerMenu {
 	public static final DeferredRegister<MenuType<?>> MENU_TYPES = DeferredRegister.create("kubejs", Registries.MENU);
-	public static final Supplier<MenuType<KubeJSMenu>> KUBEJS_MENU = MENU_TYPES.register("menu", () -> MiscPlatformHelper.get().createMenuType());
```
MaxNeedsSnacks, line 18: Why did you remove `final` here if you don't mind me asking?

ChiefArug: I was planning on also only filling this variable if serverOnly mode was off, like is done for some other registry KubeJS has, but then realised I didn't need to do that and forgot to revert it.

PR #717, opened 2023-10-03
Fix script files without a packmode not loading when packmode is not default

ChiefArug:
### Description 
By default the system treated script files without a `//packmode:` header to have a packmode of `default`, so when the packmode changed to something other than `default` those scripts didnt load.
This fixes that by defaulting to an empty string, which is already checked for later on and ignored for packmode checks (so it will load no matter the packmode).

#### Example Script 
```js
console.log('hello. i will be printed with this pr applied and packmode set to something otehr than default.')
```

#### Other details 
The only pack I know of using a packmode system is Enigmatica, and they *made their own*, likely because this didn't work. They can now use this system (cc @MuteTiefling).


ChiefArug:
This will need putting into 1.20 as well, can either be cherry picked or just done manually cause its just deleting a few characters

PR #728, opened 2023-10-22
Fix items with subtypes being put into all tabs

ChiefArug:
### Description 
Figured out here: https://discord.com/channels/303440391124942858/1165201311457890364
Subtypes fill item category method doesn't filter the tabs, so ends up adding it to all tabs...

#### Other details 
Seems like this has been a bug for a while, but gone unreported cause no one uses subtypes. This cannot be ported to 1.20 due to 1.19.3+ changing creative tabs majorly.


pietro-lopes:
Tested and works!

PR #729, opened 2023-10-22
`.DS_Store` files on macOS are spelled with capital Ds and Ss

SubordinalBlue:


MaxNeedsSnacks:
To be perfectly honest, this should probably just be equalsIgnoreCase since I think Thumbs.db is also named as such on Windows (i.e. uppercase T). Does that sound like a good idea?

SubordinalBlue:
> this should probably just be equalsIgnoreCase

Sounds good to me. Are we sure we've thought of them all? Any other common hidden file that should be added?

MaxNeedsSnacks:
Can't think of any right now, these are the only ones that I **know** are added by macOS / Windows, and Linux doesn't add any iirc

SubordinalBlue:
Wait, oops, no.

.DS_Store and Thumbs.db are _always_ capitalized that way. (Well, I'm only 90-some percent certain of that for Thumbs.db).

`equalsIgnoreCase` can't really hurt, I suppose, but it's not strictly "correct".

MaxNeedsSnacks:
With Windows ignoring capitalisation this MIGHT bite us in the ass eventually, but someone would have had to manually make a file called thumbs.db (sic) or renamed an existing Thumbs.db to that for this to happen so I'm willing to just ignore it

MaxNeedsSnacks (approved):

PR #735, opened 2023-11-04
Added Entity to RayTrace Result

pietro-lopes:
### Description 

`kjs$self().pick` returns HitResult, which calls `level().clip` that returns BlockHitResult, so it was never possible to get an Entity.

#### Example Script 
```js
ItemEvents.firstLeftClicked(event => {
  Platform.breakpoint(Object.entries(event.target))
})
```
just punch a Cow
#### Other details

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/core/EntityKJS.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/core/EntityKJS.java
@@ -304,7 +304,13 @@ public interface EntityKJS extends WithPersistentData, MessageSenderKJS, ScriptT
 	}
 
 	default RayTraceResultJS kjs$rayTrace(double distance, boolean fluids) {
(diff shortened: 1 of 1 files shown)


MaxNeedsSnacks:
Can you rebase this onto 1.19?

pietro-lopes:
For more debug context/testing
```js
PlayerEvents.tick(event => {
    if (event.player.age % 5 != 0) return
    let result = event.player.rayTrace()
    result.entity ? (event.player.statusMessage = Text.green(`Entity: ${result.entity.name.string} at ${result.entity.block.pos.toShortString()}`)) : result.block && (event.player.statusMessage = Text.aqua(`Block: ${result.block.id} at ${result.block.pos.toShortString()}`))
})
```

> Can you rebase this onto 1.19?

Yes.

pietro-lopes:
I tried something very silly, don't look at.
I'm rebasing on my pc, sec.

PR #742, opened 2023-11-14
Fixed issue where fluid render types only applied to source block.

sclark-dev:
### Description 
This fixes an issue where setting `translucent()` on any fluid only changed the render type on the source block. There are bound to be some issue with this implementation, and it does have a limitation of being unable to mix and match RenderTypes in the same fluid.

#### Example Script 
```js
StartupEvents.registry("fluid", event => {
    event.create("translucent_fluid")
        .stillTexture("kubejs:block/translucent_fluid_still")
        .flowingTexture("kubejs:block/translucent_fluid_flow")
        .bucketColor(0xFFFFFF)
        .translucent()
        .displayName("Translucent Fluid");
});
```

#### Other details


MaxNeedsSnacks:
I doubt someone would want to use separate render types for their flowing / source fluid blocks, but if it comes up, that should be easy to change nonetheless. Thanks for the commit!

MaxNeedsSnacks:
Implemented in f5fe78d / a60e3e9, gonna close this PR for now ^^

PR #758, opened 2023-12-19
Fix the regex ingredient being incredibly borked

ChiefArug:
### Description 
It will now actually serialize and deserialize using the same key, and it will actually serialize the full pattern and flags, instead of just the pattern (which before this couldn't be understood by the method used to parse it so would always be null).

I have no idea how no one has noticed this being broken yet.

#### Example Script 
```js
ServerEvents.recipes(event => {
	event.shapeless('apple', [/oak/])
})
```


ChiefArug:
I also added a null check to the pattern incase Utils.parseRegex returns null (as it can do), to provide a proper error instead of having a rubbish one in the test method.

MaxNeedsSnacks (approved):

PR #759, opened 2023-12-19
Helpful RecipeComponent extras

ChiefArug:
### Description 
This PR includes many things related to RecipeComponents:
- Adds a MappingRecipeComponent, which applies the given functions to the input before another recipe component gets it, and after the other recipe output returns a serialized output.
- Adds a SimpleMappingRecipeComponent, which uses the MappingRecipeComponent to change the keys in a json/map before it is passed to the inner component and after the inner component serializes something. This is useful for scripters when a mod uses the same format, just with different key names (like all mods and vanilla with block state inputs. see example)  
- Adds methods to RecipeComponent for the two above components. As these are mainly intended for use by scripters they are documented with @Info annotations with some documentation.
- Registers the enum component in the recipe component registry, for use by scripters. This is a dynamic component that takes one argument, `class` for the enum class.
- Adds fluidTag recipe component to the recipe component registry
- Migrates the blockstate components to default to the object form that vanilla uses, and moves the old string format to its own registry entry. Note that either will work as input in scripts, the only difference between the two is what it serializes to. It is intended that scripters use simpleMap with this component to map it to a mods needs, due to mods using different names for the block id and properties.
- Register the RegistryComponent to the recipe component register as a dynamic component that takes one component, that of the registries resource location (or a RegistryInfo from `Utils.registries`)
- Add a RemapForJS annotation to RecipeSchema#constructor as constructor is a reserved keyword in typescript, causing typing hints tools to not show these methods. For scripters it is now called addConstructor

#### Example Script 
This doesn't cover everything, but it covers most things. I do have a full set of AE2 recipes that use almost all of this stuff, but they are too long for what is meant to be a more minimal example.
```js
//requires: ae2
const $EntropyMode = Java.loadClass('appeng.recipes.entropy.EntropyMode')
const $RecipeComponentBuilder = Java.loadClass('dev.latvian.mods.kubejs.recipe.component.RecipeComponentBuilder')
const $RecipeSchema = Java.loadClass('dev.latvian.mods.kubejs.recipe.schema.RecipeSchema')

StartupEvents.recipeSchemaRegistry(event => {
    let fluidStateComponent = new $RecipeComponentBuilder(
      event.components.get('registryObject')({registry: 'fluid'}).key('id'), 
      new $MapRecipeComponent(
        event.components.get('nonBlankString')(), 
        event.components.get('nonBlankString')(), 
      false).key('properties').defaultOptional()
    )
    event.register('ae2:entropy', new $RecipeSchema(   
        new $RecipeComponentBuilder(
            event.components.get('outputBlockState')().simpleMap({ id: 'Name', properties: 'Properties' }).key('block').defaultOptional(),
            fluidStateComponent.key('fluid').defaultOptional(),
            event.components.get('outputItemArray')().key('drops').optional([])
        ).outputRole().key('output'),
        new $RecipeComponentBuilder(
            event.components.get('inputBlockState')().simpleMap({ id: 'Name', properties: 'Properties' }).key('block').defaultOptional(),
            fluidStateComponent.key('fluid').defaultOptional()
        ).inputRole().key('input'),
        event.components.get('enum')({ class: $EntropyMode }).key('mode')
    ))
})
```

#### Other details 
There are only two breaking changes here for scripters, and one of those is also for addons.
For scripters it is the constructor method name change, and the one that affects them both is the changing of the default block state component serialization state, although I doubt anyone was using that.

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/BuiltinKubeJSPlugin.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/BuiltinKubeJSPlugin.java
@@ -592,6 +592,9 @@ public class BuiltinKubeJSPlugin extends KubeJSPlugin {
 		event.register("inputBlockState", BlockStateComponent.INPUT);
 		event.register("outputBlockState", BlockStateComponent.OUTPUT);
 		event.register("otherBlockState", BlockStateComponent.BLOCK);
+		event.register("inputBlockStateString", BlockStateComponent.INPUT_STRING);
+		event.register("outputBlockStateString", BlockStateComponent.OUTPUT_STRING);
+		event.register("otherBlockStateString", BlockStateComponent.BLOCK_STRING);
 
 		event.register("ticks", TimeComponent.TICKS);
 		event.register("seconds", TimeComponent.SECONDS);
@@ -600,6 +603,7 @@ public class BuiltinKubeJSPlugin extends KubeJSPlugin {
 
 		event.register("blockTag", TagKeyComponent.BLOCK);
 		event.register("itemTag", TagKeyComponent.ITEM);
+		event.register("fluidTag", TagKeyComponent.FLUID);
 		event.register("entityTypeTag", TagKeyComponent.ENTITY_TYPE);
 		event.register("biomeTag", TagKeyComponent.BIOME);
 		event.registerDynamic("tag", TagKeyComponent.DYNAMIC);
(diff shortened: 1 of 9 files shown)


pietro-lopes:
LGTM :clap: :clap: :clap:

MaxNeedsSnacks (approved):

Yarden-zamir (approved):

PR #762, opened 2023-12-26
Added `setRandomTickCallback` for block modification

Prunoideae:
### Description 

Currently, the block modification event has a method `.setIsRandomlyTicking`, which, when set to `false`, disables the random tick of a block. However, setting this to `true` does nothing, as an originally not-ticking block will not have any logic about ticking.

This PR implements `setRandomTickCallback` for `BlockKJS` and `BlockBehaviourMixin`, making it possible for user to set custom ticking behavior for blocks that are not originally tickable.

#### Example Script 

This code snippet:

```js
// startup script
BlockEvents.modification(event => {
    event.modify('iron_block', block => {
        block.setIsRandomlyTicking(true)
        block.setRandomTickCallback(callback => {
            let { block } = callback
            block.set("copper_block")
        })
    })
})
```

Will allow the iron block to be converted to copper block when being ticked.

https://github.com/KubeJS-Mods/KubeJS/assets/24620047/d3a5f957-f2ee-4994-ba4d-26af70756c1e

#### Other details

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/core/BlockKJS.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/core/BlockKJS.java
@@ -1,6 +1,7 @@
 package dev.latvian.mods.kubejs.core;
 
 import dev.latvian.mods.kubejs.block.BlockBuilder;
+import dev.latvian.mods.kubejs.block.RandomTickCallbackJS;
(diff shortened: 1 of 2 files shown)


Untrustedlife:
Man i wish this fix was in 1.19

PR #765, opened 2024-01-02
Fixed two random problems

Prunoideae:
### Description 

1. Prevented KubeJS from crashing without clear info like in: https://gnomebot.dev/paste/1191715590537216000!latest.log. Creating a default block item as the fallback. The author of the script called `.item(item=>{...}` after `.noItem()`.
    Calling `.item` after `.noItem()` is not safe now because despite `getOrCreateItemBuilder` is called, the blockBuilder of the BlockItemBuilder is not set.
2. Modified `mods.toml` for KubeJS Forge due to crash occurred when KubeJS is clicked in the Mod Info. https://gist.github.com/Prunoideae/1ad834d4f20b82e991735b5ee56ff925.
    Where it seems that Forge in 1.19.2 requires mod icons to be in the root folder without `/`. This seems to be 1.19.2 only.


ChiefArug:
Does the fix for calling item after noItem need to be applied to fluids as well, which have both block and item/bucket for that?

Prunoideae:
It seems that fluid blocks won't have corresponding items (BucketItem≠BlockItem). The FluidBuilder does not have a `.item` method exposed but a public `FluidBucketItemBuilder` field, which should prevent users from getting random errors like in the BlockBuilder.

MaxNeedsSnacks (approved):

PR #786, opened 2024-02-08
Implement reloadable native events

LLytho:
Implement reloadable native events also added the ability to use priorities. 

Besides `NativeEvents.onEvent()` it's also possible to directly call events like `NativeEvents.clientChat()`.
If an event is client-side only, like `ScreenEvent`, we will return a dummy function to not crash

Reloading startup scripts via `/kubejs reload startup_scripts` will unregister the events on the NeoForge EventBus and will register the new listeners. Say bye to using `global` to bypass native event reloading.

Startup Scripts:
```js
NativeEvents.clientChat("low", (event) => {
	console.info("LOW PRIO: " + event.message);
});

NativeEvents.clientChat((event) => {
	console.info("NORMAL PRIO: " + event.message);
});

NativeEvents.onEvent("net.neoforged.neoforge.client.event.ClientChatEvent", event => {
	console.info("onEvent: " + event.message);
})

NativeEvents.onEvent("low", "net.neoforged.neoforge.client.event.ClientChatEvent", event => {
	console.info("onEvent low prio: " + event.message);
})

NativeEvents.mobSpawnPositionCheck((event) => {
	console.log(event.entity);
})

NativeEvents.screenInitPost((event) => {
	console.info("screenInitPost: " + event.screen);
})
```

`printAllEvents` can be used to print all possible events + their class
```js
NativeEvents.printAllEvents()
```

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/neoforge/NativeEventHandler.java
+++ b/src/main/java/dev/latvian/mods/kubejs/neoforge/NativeEventHandler.java
@@ -43,29 +43,17 @@ public class NativeEventHandler extends BaseFunction {
 				throw new IllegalStateException("Expected argument to be a function. Event syntax: " + getExampleSyntax());
 			}
 
+			var priority = (EventPriority) Context.jsToJava(cx, unknownPriority, EventPriority.class);
 			var consumer = (NativeEventConsumer) Context.jsToJava(cx, unknownFunction, NativeEventConsumer.class);
 			var securedConsumer = secure(consumer);
 			//noinspection unchecked
-			NeoForge.EVENT_BUS.addListener(getPriority(unknownPriority), false, (Class<Event>) eventClass, securedConsumer);
+			NeoForge.EVENT_BUS.addListener(priority, false, (Class<Event>) eventClass, securedConsumer);
 			consumers.add(securedConsumer);
 		} catch (Exception ex) {
 			ScriptType.STARTUP.console.error(ex.getLocalizedMessage());
 		}
 
 		return null;
-	}
-
-	private EventPriority getPriority(Object unknownPriority) {
-		if (unknownPriority instanceof EventPriority ep) {
-			return ep;
-		}
-
-		try {
-			return EventPriority.valueOf(unknownPriority.toString().toUpperCase());
-		} catch (Exception ex) {
-			ScriptType.STARTUP.console.error("Invalid priority: " + unknownPriority);
-			return EventPriority.NORMAL;
-		}
 	}
 
 	private String getExampleSyntax() {
--- a/src/main/java/dev/latvian/mods/kubejs/neoforge/NativeEventHandler.java
+++ b/src/main/java/dev/latvian/mods/kubejs/neoforge/NativeEventHandler.java
@@ -79,7 +79,7 @@ public class NativeEventHandler extends BaseFunction {
 			try {
 				consumer.accept(event);
 			} catch (Exception ex) {
-				ScriptType.STARTUP.console.error(ex.getLocalizedMessage());
+				NativeEvents.throwException("Error in native event '" + NativeEvents.NAME + "." + name + "'", ex);
 			}
 		};
 	}
--- a/src/main/java/dev/latvian/mods/kubejs/neoforge/NativeEventHandler.java
+++ b/src/main/java/dev/latvian/mods/kubejs/neoforge/NativeEventHandler.java
@@ -1,0 +1,86 @@
+package dev.latvian.mods.kubejs.neoforge;
+
+import dev.latvian.mods.kubejs.script.ScriptType;
+import dev.latvian.mods.rhino.BaseFunction;
+import dev.latvian.mods.rhino.Context;
+import dev.latvian.mods.rhino.Function;
+import dev.latvian.mods.rhino.Scriptable;
+import dev.latvian.mods.rhino.util.HideFromJS;
+import net.neoforged.bus.api.Event;
+import net.neoforged.bus.api.EventPriority;
+import net.neoforged.neoforge.common.NeoForge;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class NativeEventHandler extends BaseFunction {
+	private final List<NativeEventConsumer> consumers = new ArrayList<>();
+	private final String name;
+	private final Class<? extends Event> eventClass;
+
+	public NativeEventHandler(String name, Class<? extends Event> eventClass) {
+		this.name = name;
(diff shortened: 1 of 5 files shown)


src/main/java/dev/latvian/mods/kubejs/neoforge/NativeEvents.java
```diff
@@ -0,0 +1,147 @@
...
+		return null;
+	}
+
+	private NativeEventConsumer secure(NativeEventConsumer consumer) {
+		return event -> {
+			try {
+				consumer.accept(event);
+			} catch (Exception ex) {
+				ScriptType.STARTUP.console.error(ex.getLocalizedMessage(), ex);
+				KubeJS.LOGGER.error(ex.getLocalizedMessage(), ex);
```
ChiefArug, line 113: This should have more information about the event handler, like the errors for the callbacks in custom blocksz

src/main/java/dev/latvian/mods/kubejs/neoforge/NeoForgeEventsLookup.java
```diff
@@ -0,0 +1,206 @@
...
+	/**
+	 * Checks if given class name is invalid. Only classes which end with "Event" are valid, but also subtypes of it.
+	 *
+	 * @param fullClassName class name
+	 * @return true if class name is invalid
+	 */
+	private boolean isInvalidClassName(String fullClassName) {
+		String[] parts = fullClassName.split("\\$");
+		for (String part : parts) {
+			if (part.endsWith("Event")) {
```
ChiefArug, line 143: I don't like this as a way to check something is an event. There is no garuntee that some idiot   isn't going to end their custom events with Event (ie I could see someone making a ModIDCustomEvents class that holds a bunch of Event inner classes, excpet they are named stuff like GrowingPlant, ending up with stuff like `ModIDCustomEvents$GrowingPlant` as a plant growing event).

src/main/java/dev/latvian/mods/kubejs/neoforge/NeoForgeEventsLookup.java
```diff
@@ -0,0 +1,206 @@
...
+		}
+
+		return true;
+	}
+
+	public String formatEventName(String className) {
+		var mainParts = className.split("\\$");
+		int index = 0;
+		for (int i = mainParts.length - 1; i >= 0; i--) {
+			if (mainParts[i].endsWith("Event")) {
```
ChiefArug, line 155: Again, relying on naming conventions, although at least here its only for displaying something to the user.

ChiefArug (commented):
I also don't like that it is called `NativeEvents`, as that sounds mod loader independent. I suggest it is called something like `NeoForgeEvents` instead.

src/main/java/dev/latvian/mods/kubejs/neoforge/NeoForgeEventsLookup.java
LLytho: As this is only used for the custom calls, I only include neoforge own events and they at least follow this rule. Modded events should still go to `onEvent` imo.

src/main/java/dev/latvian/mods/kubejs/neoforge/NeoForgeEventsLookup.java
LLytho: Same as above. 

But this is just so that custom names just dont have `Event` in it. I preferred `NativeEvents.foo` instead of `NativeEvents.fooEvent`

src/main/java/dev/latvian/mods/kubejs/neoforge/NativeEvents.java
LLytho: Sounds good. Will do

LLytho:
> I also don't like that it is called `NativeEvents`, as that sounds mod loader independent. I suggest it is called something like `NeoForgeEvents` instead.

Max renamed it to `NativeEvents`, probably because it's only NeoForge from now on. I'm fine with both.

MaxNeedsSnacks:
That's right, though I should probably add ForgeEvents and NeoForgeEvents as aliases to that wrapper (although NativeEvents is the intended name for the future)

LLytho:
> That's right, though I should probably add ForgeEvents and NeoForgeEvents as aliases to that wrapper (although NativeEvents is the intended name for the future)

Will do that

ChiefArug:
> Max renamed it to `NativeEvents`, probably because it's only NeoForge from now on. I'm fine with both.

Ah, yeah. I forgot KJS was doing that.

src/main/java/dev/latvian/mods/kubejs/neoforge/NeoForgeEventsLookup.java
ChiefArug: Ah, okay. Thats good.
I wasn't sure if this was for all events or for just the 'short' version.

src/main/java/dev/latvian/mods/kubejs/neoforge/NativeEventHandler.java
```diff
@@ -0,0 +1,86 @@
...
+			ScriptType.STARTUP.console.error(ex.getLocalizedMessage());
+		}
+
+		return null;
+	}
+
+	private EventPriority getPriority(Object unknownPriority) {
+		if (unknownPriority instanceof EventPriority ep) {
+			return ep;
+		}
```
MaxNeedsSnacks, line 61: This wouldn't typewrap automatically, would it?

LLytho: Wouldn't

MaxNeedsSnacks: Then I don't quite understand why you would allow an Object here honestly?

LLytho: Because where it's being called it's an object and unknown. Thats why we just pass in the unknown object.

MaxNeedsSnacks: I don't quite get why you wouldn't then just use the builtin type wrapper here since EventPriority is an enum, but it's fine

LLytho: Oh you mean also use `jsToJava` for that. I can do that yepge

LatvianModder:
Implemented in 1.21

PR #800, opened 2024-03-02
Add support for showing the fluid tags associated with the fluid cont…

sciguyryan:
Add support for showing the fluid tags associated with the fluid contained by a BucketItem to the hand command.

### Description 

Adds support for showing the fluid tags associated with the fluid contained by a BucketItem to the KubeJS hand command output.

This could be expanded to do the same for any fluid container, but I couldn't find a way of doing this with Architectury cross-platform API. If there is a way to do this and I've missed it, feel free to let me know and I'll open a future PR to add that functionality too.

#### Example Script 
Expand the functionality of the hand command.

#### Other details 
Not applicable.

#### Example Command Output

output screenshot


ChiefArug:
It would be nice if this had a special implementation on Forge that used fluid capabilities, but that could also wait till KJS goes NeoForge only.

PR #811, opened 2024-03-21
fix event handler to persist script registry order (1.20.1) 

pietro-lopes:
### Description 
This change now persists register order as in JavaScript side.


MaxNeedsSnacks:
Superseded by fc38ad0

PR #817, opened 2024-04-07
 Implement Localization Support for CreativeTabs Titles

Qi-Month:
While working on KubeJS, I encountered an issue where I couldn't modify the titles of CreativeTabs due to their titles being hard-coded. After reviewing the code, I decided to fork the repository and implement a solution that replaces the hard-coded titles with localizable key names. This modification allows the titles of CreativeTabs to be defined in localization files, enhancing flexibility and supporting multiple languages. Additionally, I have included an en_us language file to demonstrate the localization support.

Please review this change for potential inclusion in the main branch. This update should make CreativeTabs more versatile and accessible for modders looking to support multiple languages in their Minecraft mods.

Thank you for considering this pull request.

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/item/creativetab/KubeJSCreativeTabs.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/item/creativetab/KubeJSCreativeTabs.java
@@ -9,16 +9,15 @@ import dev.latvian.mods.kubejs.registry.RegistryInfo;
 import net.minecraft.core.registries.Registries;
 import net.minecraft.network.chat.Component;
 import net.minecraft.world.item.CreativeModeTab;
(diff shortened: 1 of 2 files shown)


ChiefArug:
Can KubeJS even be translated into other languages? Its kind of a made up word, and names of things (such as companies or mods) aren't usually translated.

EvanHsieh0415:
> Can KubeJS even be translated into other languages? Its kind of a made up word, and names of things (such as companies or mods) aren't usually translated.

This can be used to rename a modpack so that the kubejs creative tab looks like it was created specifically for the modpack.

ChiefArug:
If that is the reason this is being used it might be better seeing if the creative tab title can use KubeJS' display name instead, which can be modified by scripts to change the name under custom items and the like, so that you only need to change it in one place.

PR #836, opened 2024-05-19
ParticleTypes and SoundEvents

Notenoughmail:
### Description 
Provides asset generation capabilities for `SoundEvents` and `ParticleTypes` as well as a new event for registering `ParticleProviders` for custom particles

Allows for custom sounds to work out of the box as long as scriptors have a sound file in the right auto-genned location and custom particle to work once they have a provider registered through the added event

#### Example Script 

Startup:
```js
StartupEvents.registry('sound_event', e => {
  e.create('test');
})

StartupEvents.registry('particle_type', e => {
  e.create('test');
})
```

Client:
```js
ClientEvents.particleProviderRegistry(e => {
  e.register('kubejs:test', (options, clientLevel, spriteSet, x, y, z, xSpeed, ySpeed, zSpeed) => {
    let particle = Client.customParticle(clientLevel, spriteSet, x, y, z);
    particle.gravity = 0;
    particle.friction = 1;
    particle.setParticleSpeed(0, 0, 0);
    particle.lifetime = 100;
    return particle;
  });
  // Or
  e.registerSimple('kubejs:test', (particle) => {
    particle.gravity = 0;
    particle.friction = 1;
    particle.setParticleSpeed(0, 0, 0);
    particle.lifetime = 100;
  });
  // Or
  e.registerSimple('kubejs:test'); // No modification of particle's properties
})
```

#### Other details 
A new subclass of `SimpleAnimatedParticle` was made, for use in providers, which can be made through the `Client` binding, but I'd like some input on renaming the method and/or if it would be better to move it to `ClientLevelKJS` so its use in the provider would look like `clientLevel.customParticle(...`


common/src/main/java/dev/latvian/mods/kubejs/client/SoundGenerator.java
```diff
@@ -0,0 +1,157 @@
...
+
+		private boolean replace = false;
+		@Nullable
+		private String subtitle;
+		private final List<SoundInstance> sounds = new ArrayList<>();
+
+		public SoundEntry replace(boolean b) {
+			replace = b;
+			return this;
+		}
```
ChiefArug, line 47: Should probably add an overload `replace() { return replace(true) }` like other methods of this form in KJS tend to have.

common/src/main/java/dev/latvian/mods/kubejs/core/mixin/common/ParticleEngineMixin.java
```diff
@@ -0,0 +1,18 @@
...
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@Mixin(ParticleEngine.class)
+public abstract class ParticleEngineMixin {
+
+	@Inject(method = "registerProviders", at = @At("RETURN"))
+	private void registerParticleProvidersKJS(CallbackInfo ci) {
+		ClientEvents.PARTICLE_PROVIDER_REGISTRY.post(new ParticleProviderRegistryEventJS());
+	}
+}
```
ChiefArug, line 18: Is a mixin really the best way to do this? I would think that platforms provide a way to do this.

Notenoughmail: Huh, just tested and `ParticleType`s are registered by the time `ClientLifecycleEvents#CLIENT_STARTED` posts, which I had dismissed earlier because `ClientLifecycleEvents#CLIENT_SETUP`, which states registries are initialized by then, didn't

PR #851, opened 2024-06-13
use `this.x = x` instead of `x = _x`

EvanHsieh0415:
### Description 

Use `this.x = x` instead of `x = _x` to make the parameter name more intuitive


LatvianModder:
Implemented in 1.21

PR #856, opened 2024-06-22
DRAFT: Use Locale.ROOT for converting script file names to lower case

McModknower:
### Description 
This prevents an issue faced by players with the Turkish language, as an uppercase I is changed to a lowercase dotless ı in that locale, which is not a valid character for resource names.

#### Example Script 
Any script with a capital I in the file name should trigger the bug before this pr, when using the turkish locale via `-Duser.language=tr` to force the turkish language for minecraft. Note that MultiMC automatically add `-Duser.language=en` after the java arguments from the user, masking this bug even when trying to trigger it.

#### Other details 
This is currently a draft since i want to discuss what other instances of toLowerCase/toUpperCase should be changed to include the root locale.

Edit: as a workaround players can add `-Duser.language=en` to their Java arguments.

Here is an example crash from a player that had this issue:
[crash-2024-06-01_03.45.49-fml.txt](https://github.com/user-attachments/files/15938019/crash-2024-06-01_03.45.49-fml.txt)

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/script/ScriptFileInfo.java
+++ b/src/main/java/dev/latvian/mods/kubejs/script/ScriptFileInfo.java
@@ -10,6 +10,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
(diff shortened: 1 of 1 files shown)


LatvianModder:
Implemented in 1.21

McModknower:
Is the implementation not pushed yet or in another repository?
I can't seem to find it in the main or 2004 branch.

wolfieboy09:
Lat is only doing 1.21, and won't do any other version like 1.20.1

PR #870, opened 2024-07-21
ported to ModDev

pietro-lopes:
Changed as minimum as possible.
Access Wideners -> Access Modifiers
Arch Injection interface -> NF Injection Interface

tested in dev, all runs are running fine
also tested in modpack (ATM10)

not sure if affects CI/CD :worried:


settings.gradle
```diff
@@ -1,10 +1,22 @@
 pluginManagement {
 	repositories {
+		gradlePluginPortal()
+		mavenCentral()
 		maven { url "https://maven.creeperhost.net/" }
-		maven { url "https://maven.architectury.dev/" }
+		// maven { url "https://maven.architectury.dev/" }
 		maven { url "https://maven.minecraftforge.net/" }
-		gradlePluginPortal()
+		maven {
```
Matyrobbrt, line 8: You don't need this, MDG is on GPP.

interfaces.json
Matyrobbrt, line 1: If you're only injecting one interface, you don't need an array, it can be a single string too.

pietro-lopes: Good to know! At least I didn't type it, it was already there from Arch Loom.

pietro-lopes:
@Matyrobbrt we can publish `interface.json` too, right? so addons and other mods who want to use kubejs injections can refer to it instead of manually grabbing from repo.

Matyrobbrt:
Yes. I don't think it's documented yet, but see https://github.com/neoforged/ModDevGradle/pull/89
The syntax is the same expect the block and configuration are named `interfaceInjectionData`

PR #871, opened 2024-07-23
Add some utils to `RegistryWrapper`

LLytho:
Adds some utils to the RegistryWrapper and also creates a HolderSetWrapper. 
Also changed Utils.RANDOM to use `RandomSource` which is the default random generator minecraft uses.

Example uses biomes, so it's for server events where tags already are bound.
```js
const tag = "#minecraft:is_savanna";
const registry = Registry.of("worldgen/biome");
const holders = registry.getValues(tag);
const biome = registry.getRandom();
const biomeId = registry.getId(biome);
const biome2 = holders.getRandom();
const biome2Id = registry.getId(biome2);

console.log(`All savanna biomes: ${holders.keys}`)
console.log(`Biome: ${biomeId} contains in ${tag}: ${holders.containsValue(biome)} / ${holders.contains(biomeId)}`);
console.log(`Biome 2: ${biome2Id} contains in ${tag}: ${holders.containsValue(biome2)} / ${holders.contains(biome2Id)}`);
console.log(`Biomes do exist: ${registry.containsValue(biome)} / ${registry.containsValue(biome2)}`);
console.log(registry.getValues(tag).size());
```

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/bindings/HolderSetWrapper.java
+++ b/src/main/java/dev/latvian/mods/kubejs/bindings/HolderSetWrapper.java
@@ -1,0 +1,67 @@
+package dev.latvian.mods.kubejs.bindings;
+
+import com.google.common.collect.Iterators;
+import dev.latvian.mods.kubejs.util.UtilsJS;
+import net.minecraft.core.Holder;
+import net.minecraft.core.HolderSet;
+import net.minecraft.core.Registry;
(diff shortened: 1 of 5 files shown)


ChiefArug:
RandomSource is not garunteed to be thread safe (that create method especially makes a non thread safe one that will throw if used between multiple threads), unlike Java's built in Random. This will cause issues at some point.

Issue #858, opened 2024-06-28
Error with Crossroads reagent recipes if something is done with ServerEvents.recipes' event


9thCore:
### Minecraft Version

1.20.1

### KubeJS Version

2001.6.5-build.14

### Rhino Version

2001.2.2-build.18

### Architectury Version

9.2.14

### Forge/Fabric Version

47.3.1

### Describe your issue

While running both KubeJS and Crossroads, random Crossroads reagent recipes error, causing them to not work (reagent can't be added into the recipient using the item).
Additionally, on reload, different recipes may get affected and it's random in number: could be none, could be just one, could be four. Log features multiple reload attempts to show the variance both in exact recipes and number.

Simplest script that causes this issue: `ServerEvents.recipes(e => { e.remove("minecraft:iron_pickaxe"); });`

Example of such error: `[13:58:41] [WARN] Error parsing recipe crossroads:reagents/cavorite[crossroads:reagents]: {"type":"crossroads:reagents","id":"cavorite","melting":1000,"boiling":1000,"item":"crossroads:gems/cavorite","color":"FFFF00"}: null`

Mods: Architectury, Crossroads, Essentials, KubeJS, Rhino

### Crash report/logs

https://hastebin.com/share/olehomamoq.bash

PR #895, opened 2024-08-31
(1.20.1) Make `allowAsyncStreams` affect initial recipe collection as well

9thCore:
### Description 
`allowAsyncStreams`, despite its name, does not affect all asynchronous/parallel streams. It does affect `forEachRecipe` callbacks, but not the initial recipe load done by KubeJS, which can error mods such as Crossroads (#858) which [add to a list](https://github.com/Crossroads-Development/Crossroads/blob/1.19/src/main/java/com/Da_Technomancer/crossroads/api/alchemy/ReagentManager.java#L60) when a [recipe is initialised](https://github.com/Crossroads-Development/Crossroads/blob/1.20.1/src/main/java/com/Da_Technomancer/crossroads/crafting/ReagentRec.java#L70).
Behaviour is the same if `allowAsyncStreams` is `true`, which it is by default.

#### Example Script 
Prior to this change, the simple script below would ocassionally error Crossroads Reagent recipes (see issue linked above)
```js
ServerEvents.recipes(e => { e.remove("minecraft:diamond_pickaxe"); });
```
But now, if `allowAsyncStreams` is `false`, it doesn't! ...as far as I could tell, anyway.
by being a concurrency issue it's already a roll of the dice whether it's really "fixed" or i just rolled lucky many times; I've done about 20 consecutive reloads with no such warning and all recipes still existing, whereas it took 0 (initial world load) or 1 with it being `true` to trigger a concurrency error, so I have pretty high hopes in this solution ahaha

#### Other details 
I don't think it's likely to break anything, as all it does is change the streams for recipe load to not be parallel in case `allowAsyncStreams` is false. Addons should not be using this information nor interfacing with it in any way, I do believe.

### DIFF
--- a/common/src/main/java/dev/latvian/mods/kubejs/recipe/RecipesEventJS.java
+++ b/common/src/main/java/dev/latvian/mods/kubejs/recipe/RecipesEventJS.java
@@ -377,20 +377,35 @@ public class RecipesEventJS extends EventJS {
 		var recipesByName = new HashMap<ResourceLocation, Recipe<?>>(originalRecipes.size() + addedRecipes.size());
 
 		try {
-			recipesByName.putAll(runInParallel(() -> originalRecipes.values().parallelStream()
-				.filter(RECIPE_NOT_REMOVED)
-				.map(this::createRecipe)
-				.filter(RECIPE_NON_NULL)
-				.collect(Collectors.toConcurrentMap(RECIPE_ID, RECIPE_IDENTITY, MERGE_ORIGINAL))));
+			if (CommonProperties.get().allowAsyncStreams) {
+				recipesByName.putAll(runInParallel(() -> originalRecipes.values().parallelStream()
+					.filter(RECIPE_NOT_REMOVED)
+					.map(this::createRecipe)
+					.filter(RECIPE_NON_NULL)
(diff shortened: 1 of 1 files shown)


pietro-lopes:
At mean time, should be asked to Crossroads devs to make it safe.
We had exact same problem with Terrafirma and they did it safer. So we don't have take a performance hit just because a single mod.

9thCore:
would it not be better to have consistency? i'd assumed it would affect all asynchronous streams, given its name, until i looked into kjs' source to figure out what was going on
but i guess the performance concern makes sense (even if it's but in the rare case of mods malfunctioning because of this multithreading), i'll mention the case over there for the time being

pietro-lopes:
Yes, I'm talking about *them* supporting concurrency. Should try to ask.

9thCore:
i got that part haha sorry if i wasnt clear, constructing a message right now since i need to find a concurrent alternative to `ArrayList` and im unsure how the only one i found, `CopyOnWriteArrayList`, really works

9thCore:
as shown by the pull request, issue has been solved on crossroads' end. sorry for the pr! wasn't aware tfc also had the same issue and how they fixed it

PR #904, opened 2024-09-26
Reset RegexIDFilter interner after recipes are loaded

embeddedt:
This avoids many RegexIDFilter instances remaining in memory and holding on to match caches.

In Monifactory, this saves roughly 60MB of memory.


pietro-lopes:
Can confirm, used similar strategy on ATM9 and Gravitas 2 to "save" some memory when users leave/rejoin world

MaxNeedsSnacks:
Good catch, thanks for the PR!

PR #906, opened 2024-09-27
fix: `kubejs export events` markdown url

EvanHsieh0415:
### Description 

The branch in 1.21 should be main and the path should be the path before the project structure was refactored.

https://discord.com/channels/303440391124942858/1288956688732393585

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/command/DumpCommands.java
+++ b/src/main/java/dev/latvian/mods/kubejs/command/DumpCommands.java
@@ -24,6 +24,7 @@ import java.nio.file.Files;
 public class DumpCommands {
 	private static final char UNICODE_TICK = '✔';
 	private static final char UNICODE_CROSS = '✘';
+	private static final char BRANCH_NAME = "main";
 
 	public static int events(CommandSourceStack source) {
 		var groups = EventGroups.ALL.get().map();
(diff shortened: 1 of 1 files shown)


src/main/java/dev/latvian/mods/kubejs/command/DumpCommands.java
```diff
@@ -79,9 +78,7 @@ public static int events(CommandSourceStack source) {
 
 				if (eventType.getPackageName().startsWith("dev.latvian.mods.kubejs")) {
 					builder.append('[').append(UtilsJS.toMappedTypeString(eventType)).append(']')
-						.append('(').append("https://github.com/KubeJS-Mods/KubeJS/tree/")
-						.append(KubeJS.MC_VERSION_NUMBER)
-						.append("/common/src/main/java/")
+						.append('(').append("https://github.com/KubeJS-Mods/KubeJS/tree/main/src/main/java/")
```
ChiefArug, line 81: I don't like the use of `main` as it will become outdated once KJS is ported to a new MC version.

EvanHsieh0415: I can't think of anything better I can do. Do you have any ideas?

ChiefArug: Oh wait yeah, there is no 1.21 branch, just a `main` branch. Maybe it does just need to use `main`.

EvanHsieh0415: I'll modify it when I get back to the computer. Don't merge it yet.

EvanHsieh0415: I think it can be merged?

MaxNeedsSnacks:
Oh right, this bit of code. ngl I kinda forgot it existed, because I made that before Lat started work on the new wiki and now Kubedex, so idk how long it'll stick around for now, but still, this is worth addressing... Not sure about the branch name stuff in this PR tho, I think if anything I would want to redirect people through a kubejs.com link then

PR #907, opened 2024-09-27
Adds a callback to change/override the pick block result.

Tcat2000:
### Description 
Adds a callback to change the pick block result.
#### Example Script 

```js
StartupEvents.registry("block", event => {
    event.create('example_block')
        .property(BlockProperties.AGE_4)
        .pickBlock(event => {
        	if(event.state.getValue(BlockProperties.AGE_4) == 4) event.setPickBlockItem("minecraft:stone");
        	else event.setPickBlockItem("kubejs:example_block");
        })
       ...
```

#### Other details 

I am not familiar on callbacks, so if there is a way to have 'return' in the callback return the string, instead of 'event.setPickBlockItem' that would be an improvement, but I'm not sure how to do that myself.


Tcat2000:
Ignore all of the older commits, they were reverted for this pr

common/src/main/java/dev/latvian/mods/kubejs/block/BlockBuilder.java
```diff
@@ -764,6 +770,17 @@ public BlockBuilder blockEntity(Consumer<BlockEntityInfo> callback) {
 		return this;
 	}
 
+	/**
+	 * Sets random tick callback for this black.
+	 *
+	 * @param pickBlockCallback A callback using a block container and a random.
```
ChiefArug, line 776: This javadoc is very wrong

common/src/main/java/dev/latvian/mods/kubejs/block/PickBlockCallbackJS.java
```diff
@@ -0,0 +1,36 @@
...
+import net.minecraft.core.BlockPos;
+import net.minecraft.core.registries.BuiltInRegistries;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.world.item.Item;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.BlockGetter;
+import net.minecraft.world.level.block.state.BlockState;
+
+public class PickBlockCallbackJS {
```
ChiefArug, line 12: Why is there no way to get a BlockContainerJS from here? Typically these callbacks provide one of those rather than the individual components that make it up such as BlockGetter, BlockState and BlockPos. Have a look at the other existing callbacks.

common/src/main/java/dev/latvian/mods/kubejs/block/custom/BasicBlockJS.java
```diff
@@ -31,6 +32,7 @@
 import net.minecraft.world.entity.LivingEntity;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.Items;
```
ChiefArug, line 35: Looks like some random imports got added, you can run 'Optimise Imports' in Intellij to clean them up.

ChiefArug (changes requested):

common/src/main/java/dev/latvian/mods/kubejs/block/PickBlockCallbackJS.java
Tcat2000: Alright, will do that.

PR #909, opened 2024-10-06
First attempt at reworking recipe event

MaxNeedsSnacks:
### Description 
This PR attempts to rework the recipe event such that it is less invasive, notably, we remove the unconditional cancel in our RecipeManagerMixin and instead simply modify the map of JSONs passed to the recipe manager for parsing.

### Addon Breakage
This PR *may* lead to some addons having to rework their hooks to KubeJS, especially since injectRuntimeRecipes is now being phased out.

### Additional Notes
Obviously, code cleanup is still required, including changes to logging etc. Right now this is just in a "see if things break horribly" state


src/main/java/dev/latvian/mods/kubejs/core/mixin/RecipeManagerMixin.java
```diff
@@ -31,11 +43,72 @@ public abstract class RecipeManagerMixin implements RecipeManagerKJS {
...
+			var recipesByName = new HashMap<>(byName);
+
+			KubeJSPlugins.forEachPlugin(p -> p.injectRuntimeRecipes(kjs$event, this, recipesByName));
+
+			kjs$event.finishEvent();
+
+			// make sure byType is also set correctly
+			kjs$replaceRecipes(recipesByName);
+
+			kjs$getResources().kjs$getServerScriptManager().serverData = new SyncServerDataPayload(KubeServerData.collect());
```
pietro-lopes, line 108: Now that this heavily depends on being listened to Recipe Events, this now breaks because it is null when we don't listen to Recipe Events:
https://github.com/KubeJS-Mods/KubeJS/blob/c0f67bf14c8f722709d714af5dd29608329bb839/src/main/java/dev/latvian/mods/kubejs/player/KubeJSPlayerEventHandler.java#L28

Also, is it intended that modify tooltip event should only work if listening to Recipe Events?

This is only called if listening to recipe events as:
1. `kjs$event` needs to be non-null
2. `KubeServerData.collect()` is called
3. `ItemEvents.MODIFY_TOOLTIPS#post` is finally called (this is the only reference to this post)
https://github.com/KubeJS-Mods/KubeJS/blob/c0f67bf14c8f722709d714af5dd29608329bb839/src/main/java/dev/latvian/mods/kubejs/net/KubeServerData.java#L26-L28

PR #939, opened 2025-01-17
Keybind registry and keypress events

Prunoideae:
### Description 

Added `ClientEvents.keybindRegistry`, `ClientEvents.keyDown`, and `ClientEvents.keyPressed` to register keybinds and detect if a key is down or pressed. Also added `Minecraft.kjs$isKeybindDown` to detect if a key is down in code (e.g. use in left click events).

##### startup_scripts
```js
// Event to register the keybinds
ClientEvents.keybindRegistry(event => {
    // ID of the keybind, translation key will be `key.kubejs.${id}`
    event.register('quick_buff')
        // GLFW key ID of the default key, default to -1 (Unbound)
        .keyId(event.GLFW.GLFW_KEY_A)
        // Category of the key, default to KubeJS category
        .category('key.categories.kubejs')
        // Some additional setting for key conflict and input type
        .keyConflictContext('universal')
        .inputType('keysym')
})
```

##### client_scripts
```js
// When the key is pressed, it will trigger only once
// even if you hold it for some time.
// ClientEvents.keyDown will be triggered *every tick* instead
// When the key is released, `ClientEvent.keyUp` will be triggered
ClientEvents.keyPressed('quick_buff', event => {
    if (event.player == null) return;
    event.player.playSound('minecraft:item.honey_bottle.drink', 1, 1)
    // Send to the server to notify the key is changed
    // The keypress only happens in client, so network is usually needed
    event.player.sendData('quick_buff', {})
})
```

##### server_scripts
```js
// To process the notification from client side and trigger
// the actual effect
NetworkEvents.dataReceived('quick_buff', event => {
    let inventory = event.player.inventory
    let changed = false;
    inventory.items.forEach(itemStack => {
        if (itemStack.id == "minecraft:potion") {
            let potion = itemStack.get('minecraft:potion_contents')
            let applied = false;
            potion.forEachEffect(effect => {
                if (!effect.getEffect().value().isInstantenous()) {
                    event.player.addEffect(effect)
                    applied = true;
                }
            })
            if (applied) { itemStack.shrink(1); changed = true; }
        }
    })
    if (changed) inventory.setChanged()
})
```

https://github.com/user-attachments/assets/aadf7d44-e8fb-4051-a66e-88f835b607d5

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/bindings/event/ClientEvents.java
+++ b/src/main/java/dev/latvian/mods/kubejs/bindings/event/ClientEvents.java
@@ -34,5 +34,6 @@ public interface ClientEvents {
 	EventHandler PARTICLE_PROVIDER_REGISTRY = GROUP.client("particleProviderRegistry", () -> ParticleProviderRegistryKubeEvent.class);
 	EventHandler KEYBIND_REGISTRY = GROUP.startup("keybindRegistry", () -> KeybindRegistryKubeEvent.class);
 	TargetedEventHandler<String> KEY_DOWN = GROUP.client("keyDown", () -> ClientPlayerKubeEvent.class).requiredTarget(EventTargetType.STRING);
+	TargetedEventHandler<String> KEY_UP = GROUP.client("keyUp", () -> ClientPlayerKubeEvent.class).requiredTarget(EventTargetType.STRING);
 	TargetedEventHandler<String> KEY_PRESSED = GROUP.client("keyPressed", () -> ClientPlayerKubeEvent.class).requiredTarget(EventTargetType.STRING);
 }
--- a/src/main/java/dev/latvian/mods/kubejs/bindings/event/ClientEvents.java
+++ b/src/main/java/dev/latvian/mods/kubejs/bindings/event/ClientEvents.java
@@ -5,6 +5,7 @@ import dev.latvian.mods.kubejs.client.BlockEntityRendererRegistryKubeEvent;
 import dev.latvian.mods.kubejs.client.ClientPlayerKubeEvent;
 import dev.latvian.mods.kubejs.client.DebugInfoKubeEvent;
 import dev.latvian.mods.kubejs.client.EntityRendererRegistryKubeEvent;
+import dev.latvian.mods.kubejs.client.KeybindRegistryKubeEvent;
 import dev.latvian.mods.kubejs.client.LangKubeEvent;
 import dev.latvian.mods.kubejs.client.MenuScreenRegistryKubeEvent;
 import dev.latvian.mods.kubejs.client.ParticleProviderRegistryKubeEvent;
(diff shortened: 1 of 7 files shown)


wolfieboy09:
What if there was a way to get the GLFW keys? There is `GLFW.GLFW_KEY_SEMICOLON` and one for like every other button. A class that the JS can access for the final int values would make it really useful

Prunoideae:
Added ClassWrapper to GLFW for easier access to key codes.

wolfieboy09:
Thanks Pruno! I mentioned it because not everyone knows what each number means lol

src/main/java/dev/latvian/mods/kubejs/client/KeybindRegistryKubeEvent.java
```diff
@@ -0,0 +1,74 @@
...
+	public List<KubeJSKeybinds.KubeKeybind> build() {
+		return builders.stream().map(Builder::create).toList();
+	}
+
+	public static class Builder {
+		private final String id;
+		private KeyConflictContext keyConflictContext = KeyConflictContext.UNIVERSAL;
+		private KeyModifier keyModifier = KeyModifier.NONE;
+		private InputConstants.Type inputType = InputConstants.Type.KEYSYM;
+		private int keyId = 0;
```
ChiefArug, line 37: The default should be unset (Not Bound, in the vanilla UI), rather than Left Click. (`@Nullable Integer`)

src/main/java/dev/latvian/mods/kubejs/client/KeybindRegistryKubeEvent.java
```diff
@@ -0,0 +1,74 @@
...
+import net.neoforged.neoforge.client.settings.KeyModifier;
+import org.lwjgl.glfw.GLFW;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class KeybindRegistryKubeEvent implements ClientKubeEvent {
+	private final List<Builder> builders = new ArrayList<>();
+
+	public ClassWrapper<GLFW> getGLFW() {
```
ChiefArug, line 17: This method name results in a bean of `gLFW`, not really ideal. Either the method name should be `getGlfw` or the class wrapper is a global client binding of `GLFW` (so you could do `GLFW.KEY_A`).

src/main/java/dev/latvian/mods/kubejs/bindings/event/ClientEvents.java
```diff
@@ -31,4 +32,7 @@ public interface ClientEvents {
 	TargetedEventHandler<ResourceLocation> ATLAS_SPRITE_REGISTRY = GROUP.client("atlasSpriteRegistry", () -> AtlasSpriteRegistryKubeEvent.class).requiredTarget(EventTargetType.ID);
 	TargetedEventHandler<String> LANG = GROUP.client("lang", () -> LangKubeEvent.class).requiredTarget(EventTargetType.STRING);
 	EventHandler PARTICLE_PROVIDER_REGISTRY = GROUP.client("particleProviderRegistry", () -> ParticleProviderRegistryKubeEvent.class);
+	EventHandler KEYBIND_REGISTRY = GROUP.startup("keybindRegistry", () -> KeybindRegistryKubeEvent.class);
+	TargetedEventHandler<String> KEY_DOWN = GROUP.client("keyDown", () -> ClientPlayerKubeEvent.class).requiredTarget(EventTargetType.STRING);
+	TargetedEventHandler<String> KEY_PRESSED = GROUP.client("keyPressed", () -> ClientPlayerKubeEvent.class).requiredTarget(EventTargetType.STRING);
```
ChiefArug, line 37: No key released event? I could definitely see people wanting one.

src/main/java/dev/latvian/mods/kubejs/client/KubeJSKeybinds.java
```diff
@@ -0,0 +1,71 @@
...
+import net.minecraft.client.Minecraft;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class KubeJSKeybinds {
+	private static final Map<String, KubeKeybind> registeredKeybinds = new HashMap<>();
+	private static final List<KubeKeybindState> listeningKeybinds = new ArrayList<>();
```
ChiefArug, line 15: This should be a Set (and KubeKeybindState properly implement equals and hashcode) so that multiple listeners don't result in unnecessary keybind states.

ChiefArug (changes requested):

src/main/java/dev/latvian/mods/kubejs/client/KeybindRegistryKubeEvent.java
MaxNeedsSnacks: I think a public final field of ClassWrapper<GLFW> should also work?

PR #946, opened 2025-02-01
Made container menu to be able to scale to any size

Prunoideae:
### Description 

Removed the restriction of must having 9 columns for the inventory GUI in `ServerPlayer.openInventoryGUI` and made it possible to specify the rows and columns to any size up to 9 (columns) * 6 (rows).

#### Example Script 

```js
ItemEvents.rightClicked('minecraft:diamond', event => {
    let { player } = event
    let container = new $SimpleContainer([Item.getEmpty(), Item.getEmpty(), Item.getEmpty(), Item.getEmpty(),])
    // Opens the inventory gui with the given container
    // Both width and height can be specified, first width*height in the container will be used
    // Error will be thrown if the container slots are less than width*height
    // It's also ok to use only width, or omit both width and height
    // Width and height will be automatically calculated considering the size
    player.openInventoryGUI(container, Component.translatable('item.minecraft.diamond'), 2, 2)
})
```

image

#### Other details


ChiefArug:
How does this work with vanilla/non KubeJS clients? or did the old system not work with those anyway?

Prunoideae:
The old system does not work, it has `KubeJSScreen` which is a client-side class to render the container screen. The PR did not create any new client-side classes but modify the old ones to make it functional.

Prunoideae:
And the one will work with non-KJS client is `dev.latvian.mods.kubejs.gui.chest.CustomChestMenu` which is in `kjs$openChestGUI`, the PR fixed the method `kjs$openInventoryGUI` that returns a different menu.

PR #949, opened 2025-02-03
Added explicit casting function to JavaWrapper

Prunoideae:
### Description 

Added a function `cast(Class<T>, Object obj)` that manually performs type-wrapping to any given type to avoid problems due to type erasure.

For example, `BlockState` has a such method:

```java
public <T extends Comparable<T>, V extends T> S setValue(Property<T> property, V value)
```

Where `V` is erased to `Object` at runtime, so it is impossible for Rhino to determine the correct type to wrap. Consider such script:

```js
let $CropBlock = Java.loadClass("net.minecraft.world.level.block.CropBlock")

BlockEvents.rightClicked('minecraft:wheat', event => {
    let { block: { pos, blockState }, level, hand } = event
    if (hand != "main_hand") return;

    let age = blockState.getValue($CropBlock.AGE) + 1
    level.setBlock(pos, blockState.setValue($CropBlock.AGE, age), 2)
})
```

The age is a JS value due to the `+1` operation. In Rhino before, it is directly passed into the `setBlock` and cause an error due to the number is a double, and Rhino now will report:

```
Error in 'BlockEvents.rightClicked': dev.latvian.mods.rhino.EvaluatorException: Can't find method dev.latvian.mods.rhino.CachedClassInfo.setValue(net.minecraft.world.level.block.state.properties.IntegerProperty,number).
```

both stop the original logic from working.

Thus, an explicit type wrapping method is added in this PR as `Java.cast`, this allows users to use Rhino's type wrapping to cast a value to any type needed to pass into the method.

#### Example Script 

```js
let $CropBlock = Java.loadClass("net.minecraft.world.level.block.CropBlock")
let $Integer = Java.loadClass("java.lang.Integer")

BlockEvents.rightClicked('minecraft:wheat', event => {
    let { block: { pos, blockState }, level, hand } = event
    if (hand != "main_hand") return;

    let age = blockState.getValue($CropBlock.AGE) + 1
    level.setBlock(pos, blockState.setValue($CropBlock.AGE, Java.cast($Integer, age)), 2)
})
```

So, the `age` is casted to `java.lang.Integer` and then Rhino can call the method without problem.

#### Other details 

Similar functions that require such casting exist in Neoforge / Vanilla other than BlockState properties, e.g. DataComponent and Attachment. It might be ok to implement explicit support for blockstate like for DataComponent, but I think it would be better to just let users cast things to correct types directly.

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/plugin/builtin/wrapper/JavaWrapper.java
+++ b/src/main/java/dev/latvian/mods/kubejs/plugin/builtin/wrapper/JavaWrapper.java
@@ -44,6 +44,10 @@ public interface JavaWrapper {
 		return Cast.to(cx.createInterfaceAdapter(targetClass, function));
 	}
 
+	static <T> T cast(Context cx, Class<T> targetClass, Object object) {
+		return Cast.to(cx.jsToJava(object, TypeInfo.of(targetClass)));
+	}
+
 	@Nullable
 	@HideFromJS
 	static Class<?> tryLoadClass(String className) {
--- a/src/main/java/dev/latvian/mods/kubejs/plugin/builtin/wrapper/JavaWrapper.java
+++ b/src/main/java/dev/latvian/mods/kubejs/plugin/builtin/wrapper/JavaWrapper.java
@@ -44,6 +44,7 @@ public interface JavaWrapper {
 		return Cast.to(cx.createInterfaceAdapter(targetClass, function));
 	}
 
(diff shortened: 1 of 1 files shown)


ChiefArug:
With this is the Util method for creating functional interface instances still required, or can thishjust replace it in all cases?

ChiefArug (changes requested):
Can an `@Info` annotation be added to document it please.

Prunoideae:
Ok done.

PR #956, opened 2025-02-12
Optimized GuigraphicsMixin's code

westernat:
### Description 
use WrapWithCondition instead of multiple Inject

#### Other details

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/core/mixin/GuiGraphicsMixin.java
+++ b/src/main/java/dev/latvian/mods/kubejs/core/mixin/GuiGraphicsMixin.java
@@ -1,5 +1,7 @@
 package dev.latvian.mods.kubejs.core.mixin;
 
+import com.llamalad7.mixinextras.injector.v2.WrapWithCondition;
+import com.llamalad7.mixinextras.sugar.Local;
 import dev.latvian.mods.kubejs.CommonProperties;
 import dev.latvian.mods.kubejs.client.ClientProperties;
 import dev.latvian.mods.kubejs.client.KubeJSClient;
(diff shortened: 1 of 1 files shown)


ChiefArug:
Why is this using `@Local(argsOnly = true)` rather than simply appending the enclosing methods parameters, as you can do with most injectors (including WrapWithConditions according to the wiki: https://github.com/LlamaLad7/MixinExtras/wiki/WrapWithCondition)?

pietro-lopes:
> Why is this using `@Local(argsOnly = true)` rather than simply appending the enclosing methods parameters, as you can do with most injectors (including WrapWithConditions according to the wiki: https://github.com/LlamaLad7/MixinExtras/wiki/WrapWithCondition)?

Do you have an example?
I don't see what you mean as his target is something that does not have those on the methods parameters.

ChiefArug:
> Do you have an example

I found this in GeckoLib: https://github.com/bernie-g/geckolib/blob/refs%2Fheads%2Fmain/common%2Fsrc%2Fmain%2Fjava%2Fsoftware%2Fbernie%2Fgeckolib%2Fmixin%2Fclient%2FTextureManagerMixin.java#L42-L47

pietro-lopes:
> > Do you have an example
> 
> I found this in GeckoLib: https://github.com/bernie-g/geckolib/blob/refs%2Fheads%2Fmain/common%2Fsrc%2Fmain%2Fjava%2Fsoftware%2Fbernie%2Fgeckolib%2Fmixin%2Fclient%2FTextureManagerMixin.java#L42-L47

Looks like normal to me, the invoke target contains the same args as the injected method, not extra args from parent method

ChiefArug:
Oh, I missed that the first parameter was the instance being called on and just counted the params.

src/main/java/dev/latvian/mods/kubejs/core/mixin/GuiGraphicsMixin.java
```diff
@@ -1,43 +1,24 @@
...
-	private void kjs$afterDrawSize(Font font, ItemStack stack, int x, int y, String text, CallbackInfo ci) {
-		kjs$itemSize[0] = 0;
-	}
-
-	@Inject(method = "drawString(Lnet/minecraft/client/gui/Font;Ljava/lang/String;IIIZ)I", at = @At("HEAD"), cancellable = true)
-	private void kjs$drawString(Font font, String text, int x, int y, int color, boolean dropShadow, CallbackInfoReturnable<Integer> cir) {
-		if (kjs$itemSize[0] != 0) {
-			cir.setReturnValue(KubeJSClient.drawStackSize((GuiGraphics) (Object) this, font, kjs$itemSize[0], kjs$itemSize[1], kjs$itemSize[2], color, dropShadow));
+	@WrapWithCondition(method = "renderItemDecorations(Lnet/minecraft/client/gui/Font;Lnet/minecraft/world/item/ItemStack;IILjava/lang/String;)V", at= @At(value = "INVOKE", target = "Lnet/minecraft/client/gui/GuiGraphics;drawString(Lnet/minecraft/client/gui/Font;Ljava/lang/String;IIIZ)I"))
+	private boolean kjs$drawSize(GuiGraphics instance, Font font, String text, int x, int y, int color, boolean dropShadow, @Local(argsOnly = true) ItemStack stack, @Local(argsOnly = true, ordinal = 0) int pX, @Local(argsOnly = true, ordinal = 1) int pY, @Local(argsOnly = true) String pText) {
```
ChiefArug, line 17:
```suggestion
	private boolean kjs$drawSize(GuiGraphics instance, Font font, String text, int x, int y, int color, boolean dropShadow, Font font2, ItemStack stack, int pX, int pY, String pText) {
```

ChiefArug: I'm not sure how good using the arg append method is vs using specific `@Local(argsOnly = true)` parameters, but this is what I was referring to above.

westernat: The change you're suggesting is just to make the method have one more Font parameter to be passed in

ChiefArug: Yes, but it avoids using `@Local` which in some cases can be a bad thing. I'm not sure if it is here though.

westernat: Maybe it's better now

ChiefArug: I asked some people, and it doesn't really mattter, any changes to the method are likely to break the mixin either way.

PR #957, opened 2025-02-12
Added non-conflicting methods delegating tests to `Holder.is`

Prunoideae:
### Description 

`Holder<T>` is often used in the game coding, however it contains 5 overrides of `is` method each with different parameter type, which will cause Rhino to just not work for most of the time, since many of these types are overlapping in type wrapping (e.g. `TagKey`, `ResourceKey` and `ResourceLocation`.

This PR leaves only `is(ResourceLocation)` to test if the resource location is the one holder holds, and renamed `is(Predicate<ResourceKey>)` to `test(Predicate<ResourceKey>)`, and added `isTag(ResourceLocation)` to test if the holder has a key represented by the resource location.

#### Example Script 

```js
PlayerEvents.tick(event => {
    let player = event.player
    if (player.tickCount % 80 == 0) {
        let level = event.level
        let biome = level.getBiome(player.block.pos)
        console.info(biome.is("minecraft:snowy_taiga"))
        console.info(biome.isTag("c:is_cold"))
    }
})
```

#### Other details 

- Other `is` overrides are all tagged `@HideFromJS`
- A separate `isTag` is needed instead of remapping names due to Rhino is unable to wrap the tag from type `TagKey<T>`.

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/core/mixin/HolderMixin.java
+++ b/src/main/java/dev/latvian/mods/kubejs/core/mixin/HolderMixin.java
@@ -1,39 +1,46 @@
 package dev.latvian.mods.kubejs.core.mixin;
 
+import dev.latvian.mods.kubejs.typings.Param;
+import dev.latvian.mods.kubejs.util.Cast;
+import dev.latvian.mods.rhino.util.HideFromJS;
+import dev.latvian.mods.rhino.util.RemapForJS;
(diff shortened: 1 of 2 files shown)


ChiefArug:
While we are here, can we get some @Info annotations on those? Including the one left untouched (it will also be nice to include it shadowed into there, so if it changes it will error).

PR #976, opened 2025-03-21
Fix items with subtypes being put into all tabs

jerem2772:
This PR fixes an old but previously unreported issue where items using .subtypes(...) would appear in every creative tab, including the trash slot.

The issue was caused by the fillItemCategory(...) method in BasicItemJS not properly checking whether the current creative tab matched the one the item was assigned to. As a result, NBT variants generated via .subtypes() were injected into every tab.

Seems like this has been a bug for a while, but gone unreported cause no one uses subtypes() 😅

This change ensures that variants are only added to the correct creative tab, as intended by the group(...) assignment in scripts.


pietro-lopes:
Hehe, looks like a regression
* https://github.com/KubeJS-Mods/KubeJS/pull/728

jerem2772:
Thanks for the reference!
You're right — this was indeed addressed in [#728], but that fix was made on the 1902 branch (1.19.2+), while this PR targets the 1802 branch (1.18.2).

Since I'm using .subtypes() on 1.18.2, the bug is still present in that version.
This PR reintroduces the fix specifically for the 1.18.2 line.

PR #981, opened 2025-03-25
Add text translatable fallback method

EvanHsieh0415:
### Description 

Add `.fallback(String text)` for `Text.translate()`

#### Example Script 

```js
Text.translate("text.kubejs.foo").fallback("FooBar");
Text.translatable("text.kubejs.foo").fallback("FooBar");
```

#### Other details 

This is my first submission, so I may need more suggestions to improve this feature.

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/core/ComponentKJS.java
+++ b/src/main/java/dev/latvian/mods/kubejs/core/ComponentKJS.java
@@ -77,6 +77,14 @@ public interface ComponentKJS extends Component, WithCodec, WrappedJS {
 		kjs$asIterable().forEach(action);
 	}
 
+	default MutableComponent kjs$fallback(String text) {
+    	TranslatableContents content = this.getContents();
+    	Object[] args = content.getArgs();
+    	if (args == null)
+			return Component.translatableWithFallback(new TranslatableContents(content.getKey(), text, TranslatableContents.NO_ARGS));
+    	return Component.translatableWithFallback(new TranslatableContents(content.getKey(), text, ...args));
+	}
+
 	// region ChatFormatting extensions
 	default MutableComponent kjs$black() {
 		return kjs$self().withStyle(ChatFormatting.BLACK);
--- a/src/main/java/dev/latvian/mods/kubejs/core/ComponentKJS.java
+++ b/src/main/java/dev/latvian/mods/kubejs/core/ComponentKJS.java
@@ -77,14 +77,6 @@ public interface ComponentKJS extends Component, WithCodec, WrappedJS {
 		kjs$asIterable().forEach(action);
 	}
 
-	default MutableComponent kjs$fallback(String text) {
-    	TranslatableContents content = this.getContents();
-    	Object[] args = content.getArgs();
-    	if (args == null)
-			return Component.translatableWithFallback(new TranslatableContents(content.getKey(), text, TranslatableContents.NO_ARGS));
-    	return Component.translatableWithFallback(new TranslatableContents(content.getKey(), text, ...args));
-	}
-
 	// region ChatFormatting extensions
 	default MutableComponent kjs$black() {
 		return kjs$self().withStyle(ChatFormatting.BLACK);
--- a/src/main/java/dev/latvian/mods/kubejs/plugin/builtin/wrapper/TextWrapper.java
+++ b/src/main/java/dev/latvian/mods/kubejs/plugin/builtin/wrapper/TextWrapper.java
@@ -239,6 +239,11 @@ public interface TextWrapper {
 		return Component.translatable(key, objects);
 	}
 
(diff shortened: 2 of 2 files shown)


src/main/java/dev/latvian/mods/kubejs/core/ComponentKJS.java
```diff
@@ -77,6 +77,14 @@ default void forEach(Consumer<? super Component> action) {
 		kjs$asIterable().forEach(action);
 	}
 
+	default MutableComponent kjs$fallback(String text) {
+    	TranslatableContents content = this.getContents();
+    	Object[] args = content.getArgs();
+    	if (args == null)
+			return Component.translatableWithFallback(new TranslatableContents(content.getKey(), text, TranslatableContents.NO_ARGS));
+    	return Component.translatableWithFallback(new TranslatableContents(content.getKey(), text, ...args));
```
ChiefArug, line 85: Java has no spread operator

src/main/java/dev/latvian/mods/kubejs/core/ComponentKJS.java
```diff
@@ -77,6 +77,14 @@ default void forEach(Consumer<? super Component> action) {
 		kjs$asIterable().forEach(action);
 	}
 
+	default MutableComponent kjs$fallback(String text) {
+    	TranslatableContents content = this.getContents();
```
ChiefArug, line 81: `TranslatableContents` has not been imported.

ChiefArug, line 81: `getContents()` does not return `TranslatableContents`.

ChiefArug (changes requested):
This does not compile. Please test your changes before making a PR, using the GitHub web editor to make changes isn't a great idea because you cannot test it there (and it doesn't show syntax errors).

ChiefArug:
I'm not sure I like the idea of adding this to all components, as it is only usable for translatable components and not literal/keybind/nbt/score ect components.

EvanHsieh0415:
This feature still needs more corrections, I will continue to study how to make this feature better

EvanHsieh0415:
which one is better?
1. replace from
  ```java
    @Info("Returns a translatable component of the input key, with args of the objects")
    static MutableComponent translate(String key, @Nullable String fallback, Object... objects) {
        return Component.translatableWithFallback(key, fallback, objects);
    }
  ```
  to
  ```java
    @Info("Returns a translatable component of the input key, with args of the objects")
    static MutableComponent translate(String key, Object... objects) {
        return Component.translatable(key, objects);
    }
  ```
  example:
  ```js
  Text.translate("text.kubejs.foo", "FooBar", 1, 2, 3); // with fallback("FooBar")
  Text.translate("text.kubejs.foo", null, 1, 2, 3); // no fallback
  ```
  But add null to the front of each object.
  ```diff
  - Text.translate("text.kubejs.foo", 1, 2, 3);
  + Text.translate("text.kubejs.foo", null, 1, 2, 3);
  ```
2. add a new one
  ```java
    @Info("Returns a translatable component of the input key, with args of the objects")
    static MutableComponent translate(String key, @nullable String fallback, Object... objects) {
        return Component.translatableWithFallback(key, fallback, objects);
    }
  ```
  example:
  ```js
  Text.translate("text.kubejs.foo", "FooBar", 1, 2, 3); // with fallback("FooBar") and objects(1, 2, 3)
  Text.translate("text.kubejs.foo", null, "FooBar", 1, 2, 3); // no fallback but with objects("FooBar", 1, 2, 3)
  Text.translate("text.kubejs.foo", 1, 2, 3); // no fallback
  ```
  But you have to add the `null` parameter to the front of each call where the first object is a String.
  ```diff
  Text.translate("text.kubejs.foo", 1, 2, 3); // Not need to change
  - Text.translate("text.kubejs.foo", "FooBar", 2, 3);
  + Text.translate("text.kubejs.foo", null, "FooBar", 2, 3);
  ```

ChiefArug:
I would opt to add a new method with a different name to prevent ambiguity issues. I.E. `translatableWithFallback`.

EvanHsieh0415:
圖片
It's working well.

src/main/java/dev/latvian/mods/kubejs/plugin/builtin/wrapper/TextWrapper.java
```diff
@@ -239,6 +239,16 @@ static MutableComponent translate(String key, Object... objects) {
 		return Component.translatable(key, objects);
 	}
 
+	@Info("Returns a translatable component of the input key, with args of the objects and a fallback")
```
ChiefArug, line 242: It might be worth saying 'and a fallback translation in case the client does not have one', although that is very lengthy 😕

ChiefArug (approved):
Looks good!

ChiefArug (approved):

PR #985, opened 2025-04-01
Fix default eating time in FoodBuilder.java

NatSquared:
### Description 

Fixes an apparent bug where the default eating time is comically short; _less_ than the default time for foods that are `fastToEat`, and updates them to match the Minecraft default of 1.6 seconds

#### Example Script 

```js
    event.create("sliced_bread").food((food) => {
        food.nutrition(4).saturation(0.25); // this now takes 1.6 seconds
    });

    event.create("cooked_warped_fungus").food((food) => {
         food.nutrition(2).saturation(0.25).fastToEat(); // This takes 0.8 seconds, which now matches the comment on line 77 of FoodBuilder.java
    });
```

#### Other details 
Relevant comment from line 77 of FoodBuilder.java
```java
	@Info("Sets the food is fast to eat (having half of the eating time).")
	public FoodBuilder fastToEat() {
		return eatSeconds(0.8F);
	}
```

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/item/FoodBuilder.java
+++ b/src/main/java/dev/latvian/mods/kubejs/item/FoodBuilder.java
@@ -31,7 +31,7 @@ public class FoodBuilder {
 		this.nutrition = 0;
 		this.saturation = 0;
 		this.alwaysEdible = false;
-		this.eatSeconds = 0.6F;
+		this.eatSeconds = 1.6F;
 		this.usingConvertsTo = Optional.empty();
 		this.effects = new ArrayList<>();
 	}


MaxNeedsSnacks:
Uh oh... I wonder if this bug has existed for longer than just 1.21

NatSquared:
Doesn't look like it... it looks like the FoodBuilder has been overhauled between the [1.20.4 version](https://github.com/KubeJS-Mods/KubeJS/blob/2004/src/main/java/dev/latvian/mods/kubejs/item/FoodBuilder.java) and the version seen here in 1.21.1.
The `meat` property has also been removed and `hunger` has been changed to `nutrition`, etc. (i'll probably make a PR to update the docs about it Soon™)

Issue #990, opened 2025-04-14
[EASY FIX] Doors are getting the wrong tag


Disketaa:
### Minecraft Version

1.21.1

### KubeJS Version

2101.7.1-build.181

### Rhino Version

2101.2.7-build.74

### Architectury Version

13.0.8

### Forge/Fabric Version

Neoforge 21.1.147

### Describe your issue

![Image](https://github.com/user-attachments/assets/4fa05408-80f0-419f-9bec-f7c5ee5bdd3f)

I created custom door, but for some reason KubeJS auto-inserted it in trapdoor tag, not door one

```
StartupEvents.registry('block', event => {
  event.create('tin_door', 'door')
    .soundType('iron')
    .hardness(3.0)
    .resistance(3.0)
    .requiresTool(true)
    .mapColor('terracotta_white')
})
```

wolfieboy09:
The issue persists here where Kube get's the trap doors tag: 
https://github.com/KubeJS-Mods/KubeJS/blob/1eb9f54d2609ddc9d6522e4f26a57d037abae731/src/main/java/dev/latvian/mods/kubejs/block/custom/DoorBlockBuilder.java#L35-L37

wolfieboy09:
There is `BlockTags.DOORS` and `BlockTags.WOODEN_DOORS`, so the script would have doors tag by default, and wooden doors if specified with `.wooden()` or any better method name

PR #991, opened 2025-04-15
Fix the doors builder

wolfieboy09:
### Description
Fixes the following issue: https://github.com/KubeJS-Mods/KubeJS/issues/990

#### Example Script 

```js
  event.create('unwooden_door', 'door')
    .soundType('polished_tuff')
    .hardness(3.0)
    .resistance(3.0)
    .requiresTool(true)
    .mapColor('terracotta_white')

      event.create('wooden_door', 'door')
        .soundType('polished_tuff')
        .hardness(3.0)
        .resistance(3.0)
        .requiresTool(true)
        .mapColor('terracotta_white')
        .wooden() // Tell KubeJS that this door is wooden
})
```

#### Other details
When you do `/kubejs hand`, these are the tags: 
Unwooden door: 
image
<br>
Wooden door: 
image


wolfieboy09:
Makes sense, so what would be better? for it though?

wolfieboy09:
ok

PR #1009, opened 2025-05-21
Add a command for kubejs /kjs quote_style double/single

xiaoliziawa:
### Description 

Used to choose between double quotes or single quotes when copying item ID or Tag.

Because I have found that many players find it tedious to format between double and single quotes when copying things similar to ItemId while writing KubeJS code

#### Example Script 

```mcfunction
/kubejs quote_style double
```

```mcfunction
/kubejs quote_style single
```

Use double style to display after pressing ctr+v in the editor
```js
Item.of("minecraft:diamond")
```
Use single style to display after pressing Ctrl+V in the editor
```js
Item.of('minecraft:dimaond')
```
#### Other details

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/CommonProperties.java
+++ b/src/main/java/dev/latvian/mods/kubejs/CommonProperties.java
@@ -40,6 +40,7 @@ public class CommonProperties extends BaseProperties {
 	public int defaultMaxStackSize;
 	public JsonElement creativeModeTabIcon;
 	public JsonElement creativeModeTabName;
+	public boolean useDoubleQuotes;
 
 	private CommonProperties() {
 		super(KubeJSPaths.COMMON_PROPERTIES, "KubeJS Common Properties");
@@ -59,6 +60,7 @@ public class CommonProperties extends BaseProperties {
 		startupErrorReportUrl = get("startup_error_report_url", "");
 		removeSlotLimit = get("remove_slot_limit", false);
 		defaultMaxStackSize = Math.max(0, Math.min(1_000_000_000, get("default_max_stack_size", 0)));
+		useDoubleQuotes = get("use_double_quotes", false);
 
 		creativeModeTabIcon = get("creative_mode_tab_icon", new JsonObject());
 		creativeModeTabName = get("creative_mode_tab_name", JsonNull.INSTANCE);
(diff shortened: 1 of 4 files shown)


xiaoliziawa:
Due to the ClickEvent limitation of Minecraft itself, it has to be reused
```mcfunction
kubejs hand
```
Command to retrieve new item ID and Tag

EvanHsieh0415:
How about template string style?
```js
Item.of(`xxx`)
```

xiaoliziawa:
> How about template string style?
> 
> ```js
> Item.of(`xxx`)
> ```

can

xiaoliziawa:
Now using mixin to modify the original ClickEvent, no need to input again
```mcfunction]
/kubejs hand
```
Get a new item ID or Tag

MaxNeedsSnacks:
> > IMO this should use an enum rather than a boolean to support template literal style strings as well. I also think a mixin to modify ALL click events is the wrong way to go about this as that will affect other mods as well.
> 
> If possible, perhaps we could revert back to my previous idea of having the user re-enter the /kubejs hand command

I think that would be better, yes. Sorry for letting the PR sit in limbo for too long, pretty big mental rut and uni got in the way, but I'm willing to merge this if the invasive mixin is removed

xiaoliziawa:
> > > IMO this should use an enum rather than a boolean to support template literal style strings as well. I also think a mixin to modify ALL click events is the wrong way to go about this as that will affect other mods as well.
> > 
> > 
> > If possible, perhaps we could revert back to my previous idea of having the user re-enter the /kubejs hand command
> 
> I think that would be better, yes. Sorry for letting the PR sit in limbo for too long, pretty big mental rut and uni got in the way, but I'm willing to merge this if the invasive mixin is removed

By the time I uttered this sentence, I had already reverted to using commands for switching. I believe using mixins is a very poor practice

xiaoliziawa:
> > > IMO this should use an enum rather than a boolean to support template literal style strings as well. I also think a mixin to modify ALL click events is the wrong way to go about this as that will affect other mods as well.
> > 
> > 
> > If possible, perhaps we could revert back to my previous idea of having the user re-enter the /kubejs hand command
> 
> I think that would be better, yes. Sorry for letting the PR sit in limbo for too long, pretty big mental rut and uni got in the way, but I'm willing to merge this if the invasive mixin is removed

Okay, I've reverted back to just using commands. I think mixins are a very foolish approach that leads to numerous issues

PR #1051, opened 2025-10-16
Fix error on empty ItemStack in key-pattern cleanup.

liopyu:
Throw KubeRuntimeException when a key entry resolves to an empty stack in KeyPatternCleanupPostProcessor.process. Null entries are still treated as air and stripped from the key/pattern. This prevents silent fallback from type wrappers that coerce bad inputs to ItemStack.EMPTY and surfaces misconfigurations during validation.

#### Example 

```js
ServerEvents.recipes(event => {
    // "minecraft:empty_bucket does not exist and will trigger an empty ItemStack error instead of air fallback
    event.shaped("minecraft:apple",
        [
            "XXX",
            "XOX",
            "XXX"
        ],
        {
            "X": "minecraft:acacia_boat",
            "O": "minecraft:empty_bucket"
        }
    )
})
```

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/recipe/schema/postprocessing/KeyPatternCleanupPostProcessor.java
+++ b/src/main/java/dev/latvian/mods/kubejs/recipe/schema/postprocessing/KeyPatternCleanupPostProcessor.java
@@ -3,6 +3,7 @@ package dev.latvian.mods.kubejs.recipe.schema.postprocessing;
 import com.mojang.serialization.Codec;
 import com.mojang.serialization.codecs.RecordCodecBuilder;
 import dev.latvian.mods.kubejs.KubeJS;
+import dev.latvian.mods.kubejs.error.KubeRuntimeException;
 import dev.latvian.mods.kubejs.recipe.KubeRecipe;
 import dev.latvian.mods.kubejs.recipe.RecipeKey;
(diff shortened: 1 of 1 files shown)


src/main/java/dev/latvian/mods/kubejs/recipe/schema/postprocessing/KeyPatternCleanupPostProcessor.java
```diff
@@ -41,13 +41,16 @@ public void process(RecipeValidationContext ctx, KubeRecipe recipe) {
...
-			if (entry.value() == null || component.isEmpty(Cast.to(entry.value()))) {
+			if (entry.value() == null) {
 				if (airs == null) {
 					airs = new CharArrayList(1);
 				}
-
 				airs.add(entry.key().charValue());
 				itr.remove();
+			} else if (component.isEmpty(Cast.to(entry.value()))) {
+				throw new dev.latvian.mods.kubejs.error.KubeRuntimeException(
```
wolfieboy09: It seems like your IDE put in the entire class path and not import it

pietro-lopes:
I think this is now deprecated, this commit f84e4c4c76e47c79252aa35a596bc94e929736cc will now error even before this post processing code executes

PR #1064, opened 2025-10-27
Remove remapping of Entity.hurt

lonevox:
`Entity` already has a method called `attack`, which is used when the entity attacks something. `hurt` is used to damage the entity it is called on, so they have different usages, which makes it confusing that they have the same name in kjs.


KonSola5:
From what I see in a clean NeoForge 1.21.1 devenv, `attack` is only a method of `Player`.
`LivingEntity` nor `Entity` do not have the `attack` method, neither by themselves nor in any interface those classes implement.

But `attack` method on `Player` is not remapped - will it shadow the `attack` method on `Entity`?

KonSola5:
I'd suggest remapping the method on `Entity` to `damage` instead.

MaxNeedsSnacks:
I think I'll let this be superceded by #1094, thank you for the PR tho!

PR #1068, opened 2025-11-02
Change composting recipe to the datamap way

wolfieboy09:
### Description 
In the composter block, there is a map. That map is now unused, and modifying it is useless. KubeJS originally was modifiing the map, and had TODO's to migrate but had no updates to it

TODO:
- [X] Adding items
- [x] Modifing items
- [x] Removing items

#### Example Script 

```js
ServerEvents.compostableRecipes(event => {
    event.add('minecraft:cooked_beef', 0.25) // Gives Steak a 25% chance of getting composted

    event.add('minecraft:stone', 0.33) // 33% chance for stone
    
    event.modify('minecraft:apple', 0.57, true) // Modifies dirt to 57%, and cannot be composted by villagers (third arg is optional)
    
    event.remove('minecraft:wheat') // Removes wheat from being composted
    
})
```

#### Other details 
`canVillagerCompost` defaults to false inside the codec, so the code followed that as well.

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/recipe/CompostableRecipesKubeEvent.java
+++ b/src/main/java/dev/latvian/mods/kubejs/recipe/CompostableRecipesKubeEvent.java
@@ -99,9 +99,5 @@ public class CompostableRecipesKubeEvent implements KubeEvent {
 		root.add("values", values);
 
 		generator.add(GeneratedData.json(ResourceLocation.parse("neoforge:data_maps/item/compostables.json"), () -> root));
-
-
-		System.out.println("JSON DATA");
-		System.out.println(root);
 	}
 }


ChiefArug:
If it stops the current one from being broken id imagine it would be an unbreaking change

Notenoughmail:
This feels redundant given the `#setCompostable` method in the `ServerEvents.generateData` event (which also has methods for all other NeoForge datamaps, see the `KubeDataGenerator` interface). It may be worth deprecating/removing the composting specific event and just verifying the datamap methods actually work

wolfieboy09:
So is this PR going to get closed due to https://github.com/KubeJS-Mods/KubeJS/commit/7eef8197d554e4a0b6c1a3976b0263cd552dc473?

wolfieboy09:
> Yep :)

image

PR #1083, opened 2025-12-09
Add new method to BindingRegisty for simple class naming

wolfieboy09:
### Description 
Lot's of bindings are mostly classes, and most times it's just the class name, and so this can simplify it a bit for adding a binding

#### Example Script
```java
public void registerBindings(BindingRegistry bindings) {
    bindings.add(SomeClassName.class); // This get's registered into kjs as "SomeClassName"
}
```

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/script/BindingRegistry.java
+++ b/src/main/java/dev/latvian/mods/kubejs/script/BindingRegistry.java
@@ -12,4 +12,10 @@ public record BindingRegistry(KubeJSContext context, Scriptable scope) {
 			context.addToScope(scope, name, value);
 		}
 	}
+
+	public void add(Class<?> value) {
+		if (value != null) {
+			context.addToScope(scope, value.getSimpleName(), value);
+		}
+	}
 }


MaxNeedsSnacks:
I considered doing something like this in the past, but this *kind of* introduces an antipattern of not naming your bindings properly at all, like for example, a lot of classes will have a "RecipeOutput" class and all of those bindings would then clash

KonSola5:
What about encouraging modders to create **namespaces** instead?
So let's say, instead of KubeJS Create introducing `CreateItem` and `CreateProcessingOutput` bindings, it would introduce a namespace object, `Create`. It could be a JS object or an interface with public static fields/methods.

So, instead of polluting global scope with many bindings:
```js
CreateItem.of("minecraft:stone", 0.2)
CreateProcessingOutput.of("minecraft:stone", 0.2)
```
A mod would introduce only one binding containing class references, methods, values, etc.:
```js
Create.Item.of("minecraft:stone", 0.2)
Create.ProcessingOutput.of("minecraft:stone", 0.2)
```

wolfieboy09:
That could work, but would require (maybe) a rewrite on bindings

MaxNeedsSnacks:
> What about encouraging modders to create **namespaces** instead? So let's say, instead of KubeJS Create introducing `CreateItem` and `CreateProcessingOutput` bindings, it would introduce a namespace object, `Create`. It could be a JS object or an interface with public static fields/methods.
> 
> So, instead of polluting global scope with many bindings:
> 
> ```js
> CreateItem.of("minecraft:stone", 0.2)
> CreateProcessingOutput.of("minecraft:stone", 0.2)
> ```
> 
> A mod would introduce only one binding containing class references, methods, values, etc.:
> 
> ```js
> Create.Item.of("minecraft:stone", 0.2)
> Create.ProcessingOutput.of("minecraft:stone", 0.2)
> ```

I like this a lot, actually, and think I would prefer it to something like this, so I'll be closing this PR. Let's keep discussing the new idea tho, including in this thread if you'd like

Notenoughmail:
For what it's worth, I've been using "namespaced" bindings [in KubeJS TFC](https://notenoughmail.github.io/kubejs_tfc/1.21.1/bindings/) for 2-3 years, they're trivial to do [java side](https://github.com/Notenoughmail/KubeJS-TFC/blob/1.21.1/src/main/java/io/github/notenoughmail/kubejstfc/implementation/bindings/TFCBindings.java)

KonSola5:
Nice, but still lowercase, is that a current limitation on binding casing?

Notenoughmail:
No, just personal preference. I find pascal case rather daft for this usage, especially since only two of my sub bindings even 'represent' an actual class

KonSola5:
Aha, understandable.

PR #1085, opened 2025-12-11
Server Registry Stuff

Notenoughmail:
Makes some adjustments so dealing with `ServerEvents.registry` is relatively nice.

- During the event, the `RegistryAccessContainer` has 'access' to non-static registries by making an empty registry on demand and providing that. This is needed for wrapping `Holder`s and `HolderSet`s of server registries
- Server registry builders now have their tags loaded
- Add a wrapper specifically for reference holders. Helpful for when a holder needs to be serialized as the `DeferredHolder` which is typically made cannot serialize

These changes were tested locally against [this commit](https://github.com/Notenoughmail/KubeJS-TFC/commit/df49254835b66611714e5007b3524010e2c0d869)

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/core/TagLoaderKJS.java
+++ b/src/main/java/dev/latvian/mods/kubejs/core/TagLoaderKJS.java
@@ -36,10 +36,12 @@ public interface TagLoaderKJS<T> {
 			}
 		}
 
-		hasDefaultTags |= !kjs$getResources().kjs$getServerScriptManager().serverRegistryTags.isEmpty();
+		var ssm = kjs$getResources().kjs$getServerScriptManager();
+
+		hasDefaultTags |= !ssm.serverRegistryTags.isEmpty();
 
 		if (hasDefaultTags || ServerEvents.TAGS.hasListeners(objStorage.key)) {
-			var preEvent = kjs$getResources().kjs$getServerScriptManager().preTagEvents.get(reg.key());
+			var preEvent = ssm.preTagEvents.get(reg.key());
 
 			var event = new TagKubeEvent(objStorage.key, reg);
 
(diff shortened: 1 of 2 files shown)


src/main/java/dev/latvian/mods/kubejs/holder/HolderWrapper.java
```diff
@@ -56,6 +56,25 @@ static Holder<?> wrap(KubeJSContext cx, Object from, TypeInfo param) {
...
+		if (from instanceof Holder.Reference<?> h) {
+			return h;
+		} else if (from == null) {
+			throw Context.reportRuntimeError("Can't interpret 'null' as a holder", cx);
+		}
+
+		var registry = cx.lookupRegistry(param, from);
+
+		if (from instanceof Holder<?> h) {
+			return Holder.Reference.createStandAlone(Cast.to(registry.holderOwner()), h.getKey());
```
MaxNeedsSnacks: Is createStandAlone guaranteed to be safe?

Notenoughmail: In what context? When serializing with a registry present (the intent & use case I have), yes. For retrieving a value/id from the registry itself (or the holder even), no

src/main/java/dev/latvian/mods/kubejs/core/TagLoaderKJS.java
```diff
@@ -56,6 +58,12 @@ public interface TagLoaderKJS<T> {
 				}
 			}
 
+			for (var e : kjs$getResources().kjs$getServerScriptManager().serverRegistryTags.entrySet()) {
+				for (var tag : e.getValue()) {
+					event.add(tag, new TagEventFilter.ID(e.getKey()));
+				}
```
MaxNeedsSnacks: nitpick: the SSM is probably used often enough here that it should be stored as a variable, just for readability's sake

Notenoughmail:
more-or-less, vanilla has some extra validation/safeguards by virtue of having a known set of registries and registry entries that this doesn't and the objects are actually added to those registry stubs, otherwise same principle

PR #1091, opened 2026-01-06
feat: add RandomSource overload to Utils.randomOf

EvanHsieh0415:
### Description 

Add an overload that accepts a `RandomSource`, allowing callers to use `Utils.getRandom()` or `Utils.newRandom(seed)` when invoking this method.

This makes it possible to explicitly control the random source, including using a seeded random generator when deterministic behavior is required.

#### Example Script 

```js
Utils.randomOf(Utils.getRandom(), [1, 2, 3]);
Utils.randomOf(Utils.newRandom(123456), [1, 2, 3]);
```

#### Other details


liopyu:
Maybe update the javadocs to say "random source" instead of "random"

PR #1093, opened 2026-01-27
1.21.1 -> 26.1 Update

liopyu:
# KubeJS 26.1 Primer

## 1. Overview
This branch is a **large-scale migration to NeoForge 26.1+**.

Core migration themes:
* Render/event API migration for the 26.1 client pipeline
* Item/DataComponent migration onto `ModifyDefaultComponentsEvent`
* Tool/material rework toward `ToolMaterial` + component-driven behavior
* Transfer API migration to `ResourceHandler`
* Recipe reload / encoding fixes for 26.1 registries, tags, and components
* Tag system migration away from old `TagManager` hooks
* Broad `ResourceLocation` -> `Identifier` source migration
* Temporary compatibility guards where 26.1 work is still incomplete
---
# 2. Categorized Change Breakdown

### Nature
API migration, refactor, compatibility fixes
### Key Changes
* **Client render/event migration to 26.1**
  * Item tint registration moved to `RegisterColorHandlersEvent.ItemTintSources`
  * `ItemTintFunctionWrapper` now implements `ItemTintSource`
* **Render type handling changed**
  * `ItemBlockRenderTypes.setRenderLayer()` is no longer used for generated blocks/fluids
  * Block and fluid render types are now written through `"render_type"` in generated model JSON
* Fixed / updated:
  * `KubeJSErrorScreen`
  * Key mapping registration
  * Item/block/fluid tint registration rework
---
### Nature
Major rework
### Key Changes
* Tool material logic is now centered on `ToolMaterial` plus item components such as:
  * `DataComponents.TOOL`
  * `DataComponents.WEAPON`
  * `DataComponents.ENCHANTABLE`
  * `DataComponents.REPAIRABLE`
* Added / used:
  * `MutableToolMaterial`
  * `ItemToolMaterials`
  * `ItemToolMaterialRegistryKubeEvent`
  * `KubeJSRegistries.ARMOR_MATERIAL`
* `ItemModificationKubeEvent#setTier()` now rebuilds tool-related components instead of relying on old-style direct tier mutation
* Painting variant handling now reads from:
  * `DataComponents.PAINTING_VARIANT`
* Armor material builder/registry support is present through:
  * `KubeJSRegistries.ARMOR_MATERIAL`
* Legacy compatibility code still present:
* `MutableToolTier`
* `toolTier` fields on handheld builders
* `DiggerItemBuilder`
* the startup event ID string `"toolTierRegistry"`
* an access transformer exposing `TieredItem#tier`
---
### Major Changes
* Item default component mutation now goes through:
  * `ModifyDefaultComponentsEvent`
* Item mutations operate on:
  * `DataComponentMap.Builder`
* Component removal semantics now use:
  * `set(component, null)`
  * rather than a dedicated remove call
* Item and block modification lifecycles are now split:
  * Item changes are posted during `ModifyDefaultComponentsEvent`
  * Block changes still run later during post-init / load-complete flow
* This branch includes fixes for startup crashes caused by:
  * invalid component mutation
  * mutating components too late
  * mutating bound/default components directly
* `RecordDefaults` now uses:
  * `DataComponentExactPredicate.EMPTY`
* Fluid ingredient handling no longer relies on the older `DataComponentPredicate.EMPTY` behavior
---
### Changes
* Fluid handlers moved to:
  * `ResourceHandler<FluidResource>`
* Item inventory wrappers moved to:
  * `ResourceHandler<ItemResource>`
* `FluidUtil` is now imported from:
  * `net.neoforged.neoforge.transfer.fluid.FluidUtil`
* Added:
  * `ResourceHandlerInventoryWrapper`
* `FluidTankAttachment.Wrapped` now implements:
  * `ResourceHandler<FluidResource>`
### Additional Fluid/Ingredient Notes
* `DataComponentFluidIngredient` is used for component-aware fluid matching
* Empty `FluidIngredient` / `SizedFluidIngredient` inputs are explicitly treated as unsupported
---
### Key Fixes
* Recipe JSON mutation now uses:
  * `ModifyRecipeJsonsEvent`
* `AfterRecipesLoadedKubeEvent` now rebuilds recipe maps with:
  * `RecipeMap.create()`
* Recipe reload now applies both:
  * postponed tags
  * pending new data components
  * before KubeJS recipe processing continues
* Recipe event ops now derive conditional/registry encoding from real registry access
* The server script manager is now bound to `ReloadableServerResources`
### Structural Improvements
* `ConditionalOps` is constructed from actual registry access
* Access to `ReloadableServerResources#newComponents` is wired through mixins/interfaces
* Tag-based ingredient encoding failures were fixed by applying postponed tags before recipe processing
* Component availability during recipe parsing was fixed by applying pending components early
* Custom ingredient emptiness checks were tightened so NeoForge custom ingredients do not crash KubeJS logic
### API / Encoding Updates
* `SizedIngredient` handling now uses:
  * `SizedIngredient.NESTED_CODEC`
* Recipe ingredient lookup now prefers:
  * `placementInfo().ingredients()`
* Crafting remainder access now uses:
  * `getCraftingRemainder()`
* Ingredient item iteration now uses:
  * `items()` streams
* Empty `Ingredient` / `FluidIngredient` inputs are no longer treated as valid
### Smithing
* Smithing handling is aligned to the newer split between:
  * `minecraft:smithing_transform`
  * `minecraft:smithing_trim`
* Smithing-related schema/data support is present for both recipe types
---
### Changes
* Old `TagManager`-style mixin logic was replaced with `RegistryAccess` / `HolderLookup` / `TagLoader` based hooks
* Tag access now uses:
  * `lookupOrThrow(registry).get(tag)`
* Tag reload integration now works through:
  * `TagLoader`
  * `Registry.PendingTags<?>`
  * `ReloadableServerResources#postponedTags`
* Registry tag binding now casts to:
  * `MappedRegistry`
  * because tag binding is no longer exposed from the generic registry interface
* Loaded tags are cached back into KubeJS registries during the new tag load flow
* `TagReloadContextKJS` is not used here; the active `ServerScriptManager` and registry are stored directly on the loader/mixin path instead
---
### Changes
* Added Access Transformer support for:
  * `Entity#setRot(float, float)`
* Player/entity messaging is now using:
  * `sendSystemMessage()`
  * `sendOverlayMessage()`
* The pick-block path now reconstructs `BlockHitResult` manually using a fresh ray trace in `IBlockStateExtensionMixin`
* Loot export logic uses:
  * `LootTableLoadEvent`
* Item use completion hooks currently inject into:
  * `ItemStack#finishUsingItem`
* `Explosion#getAffectedPlayers` is gone and no longer used
* `Animal#mobInteract` replacement work for `LivingEntity#eat()` is not present here
---
### Changes
* Chest/container click handling migrated from:
  * `ClickType`
  * to `ContainerInput`
* Respawn helpers were updated to the new:
  * `ServerPlayer.RespawnConfig`
  * `setRespawnPosition(respawnConfig, boolean)`
  * signature shape
* Client selected-slot sync uses:
  * `ClientboundSetHeldSlotPacket`
---
### Confirmed Project / Build Changes
* `ResourceLocation` -> `Identifier`
* Project versioning is now on:
  * `2601.7.2`
* Run configs set:
  * `terminal.ansi=true`
* Repository declarations were centralized under:
  * `settings.gradle`
  * using `dependencyResolutionManagement`
* JEI-hosting repositories were added there as part of the repo centralization
### Current Compatibility Guards / Temporary Disablement
* `validateAccessTransformers` is currently forced to:
  * `false`
* `architectury`, `JEI`, `REI`, and `EMI` compile dependencies are still commented out in `build.gradle`
* Client web/image routes in `KubeJSClientWeb.register()` are still commented out
---
### Rendering & Client
```text
RegisterShadersEvent -> RegisterRenderPipelinesEvent
RegisterColorHandlersEvent.Item -> RegisterColorHandlersEvent.ItemTintSources
ItemBlockRenderTypes.setRenderLayer() -> render_type JSON in generated models
GuiGraphics.blit() -> GuiGraphics.blit(RenderPipeline / RenderPipelines, ...)
```
---
### Entity / Messaging
```text
displayClientMessage -> sendSystemMessage
displayClientMessage -> sendOverlayMessage
Pick-block hit context -> manual ray trace + reconstructed BlockHitResult
Loot export hook -> LootTableLoadEvent
```
---
### Recipe / Ingredient
```text
SizedIngredient.FLAT_CODEC -> SizedIngredient.NESTED_CODEC
Recipe ingredient access -> placementInfo().ingredients()
ItemStack crafting remainder helpers -> getCraftingRemainder()
Ingredient#getItems() -> items() stream
AfterRecipesLoaded rebuild -> RecipeMap.create()
Recipe JSON mutation hook -> ModifyRecipeJsonsEvent
(Ingredient and FluidIngredient are no longer treated as valid when empty)
```
---
### Registry & Tags
```text
TagManager-era hooks -> RegistryAccess / HolderLookup / TagLoader mixins
lookup(registry, tag) -> lookupOrThrow(registry).get(tag)
entry.registry().bindTags() -> ((MappedRegistry) registry).bindTags()
ResourceLocation -> Identifier
KubeResourceLocation -> KubeIdentifier
```
---
### Tool / Components
```text
ItemToolTiers -> ItemToolMaterials
ItemToolTierRegistryKubeEvent -> ItemToolMaterialRegistryKubeEvent
Tier mutation -> ToolMaterial + DataComponents updates
Item mutation timing -> ModifyDefaultComponentsEvent
Direct component mutation -> DataComponentMap.Builder
remove(component) -> set(component, null)
Painting variant data -> DataComponents.PAINTING_VARIANT
```


MaxNeedsSnacks:
Adding my own changes here:
- Recipe component types have been reworked, they are now a (sort of) registry. Instance components should be stored directly, non-instance components should not use anything that relies on a registry or script manager access as a field (both registries and the script context are still available during wrap though)
- `KubeRecipe#stage` now throws an error, the recipe staging code has been removed (any addon that wants to can `@Overwrite` to hijack this method and remove the error message)
- Several fixes and improvements to type wrappers (partially forward-ported from 1.121.1)
- A bunch of legacy code has been removed
- FoodBuilder had to get reworked a little, I may replace it with entirely different wrappers since food and consumable are two different things now
- Replaced custom ingredients with the holder set equivalents where possible
- MinecraftKJS: isXXXDown was removed and replaced by vanilla's hasXXXDown

src/main/java/dev/latvian/mods/kubejs/block/entity/BlockEntityAttachment.java
```diff
@@ -20,17 +19,10 @@ default <CAP, SRC> CAP getCapability(BlockCapability<CAP, SRC> capability) {
...
-			return s.serializeNBT(registries);
-		}
-
 		return null;
 	}
 
 	default void deserialize(HolderLookup.Provider registries, @Nullable Tag tag) {
-		if (tag != null && getWrappedObject() instanceof INBTSerializable s) {
-			s.deserializeNBT(registries, tag);
-		}
```
MaxNeedsSnacks: Why is this code no-op?

src/main/java/dev/latvian/mods/kubejs/block/SeedItemBuilder.java
```diff
@@ -15,7 +15,7 @@
 import org.jetbrains.annotations.Nullable;
 
 public class SeedItemBuilder extends BlockItemBuilder {
-	public static class SeedKubeItem extends ItemNameBlockItem implements SpecialPlantable {
```
MaxNeedsSnacks, line 18: nitpick but we should check what these items do in vanilla and adjust our code accordingly

src/main/java/dev/latvian/mods/kubejs/client/highlight/HighlightRenderer.java
MaxNeedsSnacks, line 1: Commenting here to mark that this file is WIP, but seeing the state of Kubedex this isn't a blocker

src/main/java/dev/latvian/mods/kubejs/client/KubeJSClientEventHandler.java
```diff
@@ -187,40 +191,55 @@ public static void registerKeyMappings(RegisterKeyMappingsEvent event) {
...
+				// TODO: revisit this!
+				event.register(new FluidModel.Unbaked(
+					new Material(type.actualStillTexture),
+					new Material(type.actualFlowingTexture),
+					type.blockOverlayTexture != null ? new Material(type.blockOverlayTexture) : null,
+					null
+				), b.get());
+			}
+		}
 	}
```
MaxNeedsSnacks, line 222: todo's already there, but yeah we need to revisit a lot of the item, block and fluid model stuff

src/main/java/dev/latvian/mods/kubejs/command/InformationCommands.java
```diff
@@ -29,19 +30,19 @@ private static Component copy(String s, ChatFormatting col, Component info) {
...
 
 	public static int hand(ServerPlayer player, InteractionHand hand) {
 		player.sendSystemMessage(Component.literal("Item in hand:"));
 		var stack = player.getItemInHand(hand);
-		var holder = stack.getItemHolder();
-		var itemRegistry = player.server.registryAccess().registry(Registries.ITEM).orElseThrow();
-		var blockRegistry = player.server.registryAccess().registry(Registries.BLOCK).orElseThrow();
-		var fluidRegistry = player.server.registryAccess().registry(Registries.FLUID).orElseThrow();
-		var tabRegistry = player.server.registryAccess().registry(Registries.CREATIVE_MODE_TAB).orElseThrow();
+		var holder = Holder.direct(stack.getItem());
```
MaxNeedsSnacks: Direct holder??

src/main/java/dev/latvian/mods/kubejs/command/InformationCommands.java
```diff
@@ -29,19 +30,19 @@ private static Component copy(String s, ChatFormatting col, Component info) {
...
+		var itemRegistry = BuiltInRegistries.ITEM;
+		var blockRegistry = BuiltInRegistries.BLOCK;
+		var fluidRegistry = BuiltInRegistries.FLUID;
+		var tabRegistry = BuiltInRegistries.CREATIVE_MODE_TAB;
```
MaxNeedsSnacks: builtin regs should be avoided here

src/main/java/dev/latvian/mods/kubejs/component/ComponentFunctions.java
```diff
@@ -150,27 +152,51 @@ public interface ComponentFunctions {
...
+		kjs$override(
+			DataComponents.CUSTOM_MODEL_DATA,
+			new CustomModelData(
+				List.of(),
+				List.of(),
+				List.of(),
+				List.of(data)
+			)
+		);
 	}
```
MaxNeedsSnacks: TODO: maybe introduce other kinds of custom model data?

src/main/java/dev/latvian/mods/kubejs/component/ComponentFunctions.java
```diff
@@ -182,17 +208,39 @@ public interface ComponentFunctions {
...
+			return;
+		}
+
+		var id = tag.getString("id");
+		if (id.isEmpty()) {
+			throw new IllegalArgumentException("ENTITY_DATA tag must contain non-empty \"id\"");
+		}
+
+		var type = EntityType.byString(id.get()).orElseThrow(() -> new IllegalArgumentException("Unknown entity id: " + id));
+		kjs$override(DataComponents.ENTITY_DATA, TypedEntityData.of(type, tag));
```
MaxNeedsSnacks: would a type wrapper for TypedEntityData be a good idea here?

src/main/java/dev/latvian/mods/kubejs/core/mixin/AbstractClientPlayerMixin.java
```diff
@@ -13,6 +13,6 @@
 @RemapPrefixForJS("kjs$")
 public abstract class AbstractClientPlayerMixin extends Player implements ClientPlayerKJS {
 	public AbstractClientPlayerMixin(Level level, BlockPos blockPos, float f, GameProfile gameProfile) {
```
MaxNeedsSnacks, line 15: not sure if mixin needs these constructor params to match or not

MaxNeedsSnacks (changes requested):
Only got through about a fifth of the codebase so far (in addition to working on fixing things myself, here are a couple of notes)

Also, general note to self: Resource handlers likely need to be reworked

src/main/java/dev/latvian/mods/kubejs/block/entity/BlockEntityAttachment.java
liopyu: With the removal of INBTSerializable Neoforge is moving towards Codec encoding/decoding rather than manual serialization, this no-op removal will work for our purposes as we already handle manual serialization on our block entity end so just leaving this as no-op is a valid option. All that being said I'll go ahead and switch us over to Codec instead as it's just way better anyways and we can avoid deprecation implications in the future.

src/main/java/dev/latvian/mods/kubejs/command/InformationCommands.java
liopyu: Changed to use stored type holder

src/main/java/dev/latvian/mods/kubejs/command/InformationCommands.java
liopyu: Changed to use registryAccess + lookupOrThrow

MaxNeedsSnacks:
Additional TOOD: Component modification looks like it'll probably be going through changes on the Neo side as well; I'm not sure whether to comment it out entirely though or try and change our modification event. Changing out the filter to be an Item predicate instead of ItemStack is the easy part, the bigger problem is with components themselves not being initialised yet so we can't access the item's original component map, which is something players may have come to expect or rely on

src/main/java/dev/latvian/mods/kubejs/block/entity/BlockEntityAttachment.java
liopyu: We'll persist CustomCapabilityAttachment through an Optional Codec to its Factory record without changing the interface. this would be the least invasive change we could make for this to account for the no-op not handling CustomCapabilityAttachment anymore (and probably the best in practice until we're ready to rip out our own Attachment registries in favor of Neo's new [AttachmentType ](https://github.com/neoforged/NeoForge/blob/be14d7be69f8afa62af942d5864a7155fedcdd49/src/main/java/net/neoforged/neoforge/attachment/AttachmentType.java#L38)registry)

MaxNeedsSnacks: "until we're ready" is now; whatever breaking changes we need to make we just make

liopyu: say less

src/main/java/dev/latvian/mods/kubejs/component/ComponentFunctions.java
liopyu: added a helper method and enum setter to cover all DataComponent types
working example: 
```javascript
ItemEvents.rightClicked(event => {
    let stack = event.item
    stack.setCustomModelData('float', [1.0, 2.5])
    stack.setCustomModelData('boolean', [true, false])
    stack.setCustomModelData('string', ['variant_a'])
    stack.setCustomModelData('color', [0xFF0000, 0x00FF00])
    let cmd = stack.getCustomModelData()
    console.log("CMD: " + cmd) // CMD: CustomModelData[floats=[1.0, 2.5], flags=[true, false], strings=[variant_a], colors=[16711680, 65280]]
})
```

src/main/java/dev/latvian/mods/kubejs/core/mixin/AbstractClientPlayerMixin.java
liopyu: Mixins requires the constructor signature to match the superclass so it can generate valid bytecode, even though the constructor body is never actually called it's a compile time requirement

src/main/java/dev/latvian/mods/kubejs/component/ComponentFunctions.java
liopyu: In my opinion i think the current TypedEntityData is covering this certain use case and a separate type/enum wouldn't really add anything here because it's only one operation and the validation is pretty straight forward with id. Maybe if we expanded this in the future

src/main/java/dev/latvian/mods/kubejs/block/entity/BlockEntityAttachment.java
liopyu: This should be fixed as of [a063e69](https://github.com/KubeJS-Mods/KubeJS/pull/1093/commits/a063e69dd53fca95b2b88bae317bae0949b3faae)

src/main/java/dev/latvian/mods/kubejs/component/ComponentFunctions.java
liopyu: i actually am removing setCustomModelData(int) too because thats now covered by the 'color' override

src/main/java/dev/latvian/mods/kubejs/block/SeedItemBuilder.java
liopyu: Added a `seedBuilder.villagerCanPlant(true/false)` override to start, defaulting to true

src/main/java/dev/latvian/mods/kubejs/core/mixin/AbstractClientPlayerMixin.java
MaxNeedsSnacks: that's not what i meant; that said, mixin doesn't seem to complain about the extra parameters so it's fine

src/main/java/dev/latvian/mods/kubejs/client/KubeJSClientEventHandler.java
liopyu: I've updated it to Neoforge's new fluid API adding a `FluidTintFunctionWrapper`as well that adds neoforge's new world context to fluid tinting

liopyu: I will also look into item which will probably be the same as fluid/block modelling depending on what's available, items for the most part are inside the player's inventory which wouldnt have the world context but it's something i'll review nonetheless

MaxNeedsSnacks (changes requested):
The last set of changes (namely the error screen and custom menu changes) are out of scope for a port like this, please revert them and make another PR just for them

src/main/java/dev/latvian/mods/kubejs/core/CustomIngredientKJS.java
```diff
@@ -2,12 +2,12 @@
...
+import dev.latvian.mods.kubejs.plugin.builtin.wrapper.ItemWrapper;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.crafting.Ingredient;
 import net.neoforged.neoforge.common.crafting.ICustomIngredient;
 import org.jetbrains.annotations.ApiStatus;
 
-import java.util.stream.Collector;
 import java.util.stream.Stream;
 
 public interface CustomIngredientKJS extends ItemPredicate {
```
MaxNeedsSnacks, line 15: needs reworking (just like all of ItemPredicate tbh), getItems() doesn't exist anymore, it's all slot displays all the way down

src/main/java/dev/latvian/mods/kubejs/core/FluidIngredientKJS.java
```diff
@@ -35,20 +34,39 @@ default boolean matches(RecipeMatchContext cx, FluidStack s, boolean exact) {
...
+			for (var holder : fluids) {
+				if (in.test(new FluidStack(holder, probeAmount))) {
+					return true;
+				}
+			}
+		} catch (Exception ex) {
+			throw new KubeRuntimeException("Failed to test sized fluid ingredient " + in, ex);
+		}
 		return false;
 	}
```
MaxNeedsSnacks, line 73: I'm not that well-versed with ReplacementMatches anymore (we ***really*** need documentation, but resolving these just to test if they overlap at all feels... wrong)

We might want to look into reworking how replacement works after this PR

src/main/java/dev/latvian/mods/kubejs/core/RecipeHolderKJS.java
```diff
@@ -77,12 +88,18 @@ default boolean replaceInput(RecipeScriptContext cx, ReplacementMatchInfo match,
```
MaxNeedsSnacks, line 78: This may be able to use the ingredients from the placement info instead? Either way, probably not good to hardcode only Shaped and Shapeless here

src/main/java/dev/latvian/mods/kubejs/core/RecipeHolderKJS.java
```diff
@@ -77,12 +88,18 @@ default boolean replaceInput(RecipeScriptContext cx, ReplacementMatchInfo match,
...
+			}
+		} else if (recipe instanceof ShapelessRecipe shapeless) {
+			if (match.match() instanceof ItemMatch m) {
+				var result = shapeless.result.create();
+				return result != null && result != ItemStack.EMPTY && !result.isEmpty() && m.matches(cx, result, match.exact());
+			}
 		}
-
 		return false;
 	}
```
MaxNeedsSnacks, line 97: This one's harder because there's no stateless result anymore... We may need to rethink a lot of this down the line, like I said

src/main/java/dev/latvian/mods/kubejs/core/ServerPlayerKJS.java
```diff
@@ -138,14 +142,30 @@ public interface ServerPlayerKJS extends PlayerKJS {
 		}
 	}
 
+	default boolean kjs$hasPermission(int i) {
+		var permission = new Permission.HasCommandLevel(PermissionLevel.byId(i));
+		return kjs$self().permissions().hasPermission(permission);
+	}
```
MaxNeedsSnacks, line 184: TODO (post-port): more granular permission support?

MaxNeedsSnacks (changes requested):

src/main/java/dev/latvian/mods/kubejs/core/RecipeHolderKJS.java
liopyu: Should use placement info instead of hard coded shaped/shapeless as of [53fa88b](https://github.com/liopyu/KubeJS/commit/53fa88bbfca426943c16df6bbc1a1cb7c8c52a6b)

src/main/java/dev/latvian/mods/kubejs/registry/BuilderBase.java
```diff
@@ -52,18 +53,27 @@ public T transformObject(T obj) {
...
 	}
 
+	@HideFromJS
+	public T getOrCreate() {
+		if (object == null) {
+			createTransformedObject();
+		}
+		return object;
+	}
+
```
MaxNeedsSnacks: not a fan of this eager creation just for armour materials

src/main/java/dev/latvian/mods/kubejs/registry/CodecRegistryBuilder.java
```diff
@@ -0,0 +1,72 @@
...
+import com.google.gson.JsonObject;
+import com.mojang.serialization.Codec;
+import com.mojang.serialization.DataResult.Error;
+import com.mojang.serialization.DynamicOps;
+import dev.latvian.mods.kubejs.error.KubeRuntimeException;
+import dev.latvian.mods.kubejs.typings.Info;
+import dev.latvian.mods.kubejs.util.JsonUtils;
+import net.minecraft.resources.Identifier;
+
+public class CodecRegistryBuilder<T> extends BuilderBase<T> {
```
MaxNeedsSnacks, line 13: do you have some test cases for this?

MaxNeedsSnacks:
also a blocker: i need to update or CI workflows; probably gonna use the almostreliable ones

src/main/java/dev/latvian/mods/kubejs/registry/CodecRegistryBuilder.java
liopyu: Here's one of my server script tests from this one: 
```javascript
// run `/damage Dev 10 kubejs:example` to test
ServerEvents.registry('minecraft:damage_type', event => {
	event.create('kubejs:example').merge({
		message_id: 'example',
		scaling: 'never',
		exhaustion: 0.1,
		effects: 'hurt',
		death_message_type: 'default'
	})
})
```
Another one for the road:
image

```javascript
ServerEvents.registry('minecraft:banner_pattern', event => {
	event.create('kubejs:zigzag').merge({
		asset_id: 'minecraft:flower',
		translation_key: 'block.minecraft.banner.kubejs.zigzag'
	})
})
```
This event does not update on /reload but will on world rejoin.

src/main/java/dev/latvian/mods/kubejs/registry/BuilderBase.java
liopyu: ok then we can redesign armor items to store a lazy reference like Identifier/ResourceKey and resolve later

liopyu: you can count this as resolved, this was for custom equipment assets registry which i confused for another registry when initially porting. We can always revisit this registry through the client events or something later

liopyu: As for ArmorMaterialBuilder i just realized it'd work better as some helper function rather than an actual builder because all its defining values are in the constructor anyways

PR #1094, opened 2026-01-30
Entity-related changes

KonSola5:
# Description 
This PR adds changes related to entities, namely:
- Changed the semantics of `EntityKJS#kjs$getUsername`, since hmm, what's the username of an entity?
- Restored `Entity#getScoreboardName` to its original form.
- Restored `Entity#clearFire` and remapped `Entity#extinguishFire` to `extinguish`.
- Changed the remap of `Entity#hurt` from `attack` to `damage`. The original `attack` name still exists on `EntityKJS`, but it's marked as `@Deprecated`. Addresses what #1064 tried to do, while still creating a more descriptive name.
- Changed remaps of various `distanceTo` and `distanceToSqr` to be consistent with each other, while not being ambiguous with each other. All old remaps stay and are marked as `@Deprecated`.

| Original name (in `Entity`)                   | Old remap / name                            | New remap / name                    |
| --------------------------------------------- | ------------------------------------------- | ----------------------------------- |
| `distanceToSqr(double x, double y, double z)` | `getDistanceSq`                             | `distanceToSqr` (reverted)          |
| `distanceToSqr(Vec3 position)`                | `distanceToSqr`                             | `distanceToSqr` (stays as-is)       |
| `distanceToSqr(Entity entity)`                | `distanceToEntitySqr`                       | `distanceToEntitySqr` (stays as-is) |
| -                                             | `getDistanceSq(BlockPos pos)`               | `distanceToBlockSqr`                |
| -                                             | `getDistance(double x, double y, double z)` | `distanceTo`                        |
| -                                             | -                                           | `distanceTo(Vec3 position)`         |
| `distanceTo(Entity entity)`                   | `distanceToEntity`                          | `distanceToEntity` (stays as-is)    |
| -                                             | `getDistance(BlockPos pos)`                 | `distanceToBlock`                   |

- Changed the name of `EntityKJS#kjs$getTeamId` to `EntityKJS#kjs$getTeamName`. Again, old name exists, but `@Deprecated`.
- Changed `EntityKJS#kjs$getProfile` to return `null` in case of entities that aren't players - game profiles only make sense on players.
- Removed `EntityKJS#kjs$playSound` - changes nothing, because `Entity` already has `playSound` which respects the entity's silence state.
- Changed the `EntityKJS#kjs$teleportTo` method overloads to no longer use command stack, nor execute the logic from `TeleportCommand` (`/tp` command). Rationale:
  - `teleportTo` should not be just an alias for a command
  - NeoForge events should not affect the outcome of the teleport - the inversion of control here is simply not in place here, as you already are in control of your script, so losing control due to a random NeoForge teleport **command** event listener would create surprising behavior.
- Also, the errors thrown by `EntityKJS#kjs$teleportTo` are more descriptive - they now actually tell what's the problem.
- Added the `EntityKJS#kjs$teleportToLevel` method - needed because the type wrapper can't discern between `ServerLevel` and `ResourceLocation` - avoids unnecessary ID lookup if you already have a reference to a server level - this isn't hard to do.
- Added the `EntityKJS#kjs$isItem` method - checks whether the entity is an item entity. (Now you should reconsider what happens when both `isItem` and `getItem` exist at once...)
- Added `@Info` documentations in various places.
- `@HideFromJS`'d various `kjs$self` methods - these are useless in JS, they are more useful in Java to get the appropriate `this` for method calls.
- Added a new annotation interface: `@ThisIs`, which can help in creation of type guards for type dumping mods.
- Changed some single letter variable names in method descriptors to be... more descriptive, in case type dumping mods can read them.

## Server script
Adds a `/test` command.
```js
/** @import { $MinecraftServer } from "net.minecraft.server.MinecraftServer" */

ServerEvents.commandRegistry((event) => {
  event.register(
    event.commands.literal("test").executes((ctx) => {
      const things = {
        player: ctx.source.player,
        server: ctx.source.server,
        level: ctx.source.level,
      };
      global.testFunction(things);
      return 1;
    }),
  );
});
```

## Startup script (so it can be reloaded quickly)
```js
// @ts-check
/** @import { $MinecraftServer } from "net.minecraft.server.MinecraftServer" */
/** @import { $ServerPlayer } from "net.minecraft.server.level.ServerPlayer" */
/** @import { $ServerLevel } from "net.minecraft.server.level.ServerLevel" */
/** @import { $Entity } from "net.minecraft.world.entity.Entity" */

/** @param {{player: $ServerPlayer, server: $MinecraftServer, level: $ServerLevel}} things */
global.testFunction = function (things) {
  const { player, server, level } = things;

  player.setPositionAndRotation(0.5, -60, 0.5, 0, 0);

  /** @type {[$Entity, $Entity]} */
  const armorStands = [null, null];
  const playerPos = new Vec3d(player.x, player.y, player.z);
  const steps = {
    "Spawning armor stands..."() {
      armorStands[0] = level.createEntity("minecraft:armor_stand");
      armorStands[0].moveTo(playerPos.add([0, 0, 5]));
      armorStands[0].spawn();

      armorStands[1] = level.createEntity("minecraft:armor_stand");
      armorStands[1].moveTo(playerPos.add([3, 0, 5]));
      armorStands[1].spawn();
    },
    "Distance between stands is:"() {
      player.tell(String(armorStands[0].distanceToEntity(armorStands[1])));
      player.tell(`Square: ${armorStands[0].distanceToEntitySqr(armorStands[1])}`);
    },
    "Distance between block I'm standing is:"() {
      player.tell("Stand in front:");
      player.tell(String(armorStands[0].distanceToBlock(player.onPos)));
      player.tell(`Square: ${armorStands[0].distanceToBlockSqr(player.onPos)}`);
      player.tell("Stand on the side:");
      player.tell(String(armorStands[1].distanceToBlock(player.onPos)));
      player.tell(`Square: ${armorStands[1].distanceToBlockSqr(player.onPos)}`);
    },
    "Distance between center stand and (0, -56, 0) is:"() {
      player.tell(String(armorStands[0].distanceTo(0, -56, 0)));
      player.tell(`Square: ${armorStands[0].distanceToSqr(0, -56, 0)}`);
      player.tell("Vec3 version...");
      player.tell(String(armorStands[0].distanceTo([0, -56, 0])));
      player.tell(`Square: ${armorStands[0].distanceToSqr([0, -56, 0])}`);
    },
    "The profile of armor stand shall be null"() {
      player.tell(String(armorStands[0].profile));
    },
    "Let's make the armor stand on the side prettier."() {
      armorStands[1].customName = Text.ofString("Lilac").bold().color("#D891EF");
      armorStands[1].customNameVisible = true;
    },
    "Usernames:"() {
      player.tell(`Front: ${armorStands[0].username}`);
      player.tell(`Side: ${armorStands[1].username}`);
    },
    "Oink:"() {
      armorStands[0].playSound("minecraft:entity.pig.ambient");
    },
    "Moo:"() {
      armorStands[1].playSound("minecraft:entity.cow.ambient");
    },
    "Teleport to level:"() {
      const { x, y, z, yaw, pitch } = armorStands[0];
      armorStands[0].teleportToLevel(server.overworld(), x, y + 5, z, yaw + 90, pitch);
    },
    "Teleport:"() {
      const { x, y, z, yaw, pitch } = armorStands[1];
      armorStands[1].teleportTo("minecraft:overworld", x, y + 4, z, yaw - 90, pitch);
    },
    "Setting the front armor stand on fire:"() {
      armorStands[0].remainingFireTicks = 60;
    },
    "Extinguishing it:"() {
      armorStands[0].extinguish();
    },
  };

  /**
   * @param {ArrayIterator<[string, () => void]>} iterator
   */
  function doAfter40Ticks(iterator) {
    server.scheduleInTicks(40, () => {
      const nextStep = iterator.next();
      if (nextStep.done) return;
      /** @type {[string, () => void]} */
      const [name, callback] = nextStep.value;
      player.tell("");
      player.tell(name);
      callback();
      doAfter40Ticks(iterator);
    });
  }

  const stepEntries = Object.entries(steps)[Symbol.iterator]();
  server.runCommandSilent("/kill @e[type=minecraft:armor_stand]");
  doAfter40Ticks(stepEntries);
};
```

## Other details 
Added a new documentation annotation interface: `@ThisIs`.
Type dumping mods such as ProbeJS can use it to create type guards.

So, let's say, currently there's a method on Entity called:
```ts
declare class Entity {
  isPlayer(): boolean
  get player(): boolean
}
```
If the boolean-returning method is annotated as `@ThisIs(Player.class)`, the return type may be dumped instead as:
```ts
declare class Entity {
  isPlayer(): this is Player
  get player(): boolean
}
```
With this type guard in place, TS will automatically treat the instance as instance of the specified type inside conditional branches:
```ts
declare const entity: Entity;
if (entity.isPlayer()) {
  // give is only on players, and with this change TS will suggest it!
  entity.give("2x minecraft:apple");
}
```

> [!CAUTION]
> Do not emit type guards on beans!
> ```ts
> declare class Entity {
>   isPlayer(): this is Player
>   get player(): this is Player // ❌ invalid syntax!
> }
> ```

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/core/ClientPlayerKJS.java
+++ b/src/main/java/dev/latvian/mods/kubejs/core/ClientPlayerKJS.java
@@ -1,20 +1,28 @@
 package dev.latvian.mods.kubejs.core;
 
 import dev.latvian.mods.kubejs.player.PlayerStatsJS;
+import dev.latvian.mods.kubejs.typings.Info;
+import dev.latvian.mods.kubejs.typings.ThisIs;
 import dev.latvian.mods.kubejs.util.NotificationToastData;
+import dev.latvian.mods.rhino.util.HideFromJS;
 import dev.latvian.mods.rhino.util.RemapPrefixForJS;
 import net.minecraft.client.player.AbstractClientPlayer;
+import net.minecraft.client.player.LocalPlayer;
 import net.minecraft.nbt.CompoundTag;
 import org.jetbrains.annotations.Nullable;
 
 @RemapPrefixForJS("kjs$")
 public interface ClientPlayerKJS extends PlayerKJS {
 	@Override
+	@HideFromJS
 	default AbstractClientPlayer kjs$self() {
 		return (AbstractClientPlayer) this;
 	}
 
-	default boolean isSelf() {
+	@Override
+	@ThisIs(LocalPlayer.class)
+	@Info("Checks, whether the entity is a reference to yourself - that is - the client player you are controlling.")
+	default boolean kjs$isSelf() {
 		return false;
 	}
 
--- a/src/main/java/dev/latvian/mods/kubejs/core/ClientPlayerKJS.java
+++ b/src/main/java/dev/latvian/mods/kubejs/core/ClientPlayerKJS.java
@@ -7,7 +7,6 @@ import dev.latvian.mods.kubejs.util.NotificationToastData;
 import dev.latvian.mods.rhino.util.HideFromJS;
 import dev.latvian.mods.rhino.util.RemapPrefixForJS;
 import net.minecraft.client.player.AbstractClientPlayer;
-import net.minecraft.client.player.LocalPlayer;
 import net.minecraft.nbt.CompoundTag;
 import org.jetbrains.annotations.Nullable;
 
@@ -19,11 +18,10 @@ public interface ClientPlayerKJS extends PlayerKJS {
 		return (AbstractClientPlayer) this;
 	}
 
-	@Override
-	@ThisIs(LocalPlayer.class)
-	@Info("Checks, whether the entity is a reference to yourself - that is - the client player you are controlling.")
-	default boolean kjs$isSelf() {
-		return false;
+	@ThisIs(AbstractClientPlayer.class)
+	@Info("Checks if the entity is a client-side player.")
+	default boolean kjs$isClientPlayer() {
+		return true;
 	}
 
 	@Override
--- a/src/main/java/dev/latvian/mods/kubejs/core/EntityKJS.java
+++ b/src/main/java/dev/latvian/mods/kubejs/core/EntityKJS.java
@@ -1,12 +1,14 @@
 package dev.latvian.mods.kubejs.core;
 
 import com.mojang.authlib.GameProfile;
-import com.mojang.brigadier.exceptions.CommandSyntaxException;
 import dev.latvian.mods.kubejs.entity.KubeRayTraceResult;
 import dev.latvian.mods.kubejs.level.LevelBlock;
 import dev.latvian.mods.kubejs.player.EntityArrayList;
 import dev.latvian.mods.kubejs.script.ScriptType;
 import dev.latvian.mods.kubejs.script.ScriptTypeHolder;
+import dev.latvian.mods.kubejs.typings.Info;
+import dev.latvian.mods.kubejs.typings.Param;
+import dev.latvian.mods.kubejs.typings.ThisIs;
 import dev.latvian.mods.kubejs.util.UtilsJS;
 import dev.latvian.mods.rhino.util.HideFromJS;
 import dev.latvian.mods.rhino.util.RemapPrefixForJS;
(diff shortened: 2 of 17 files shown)


KonSola5:
Additional problems to resolve:
- In `EntityArrayList`, `oneFilter` breaks after matching one entity
- In `EntityArrayList`, `filter` can't be used easily with KubeJS because passing a JS array of functions to a `List<Predicate<Entity>>` will only cause the array to be converted into a `List`, functions inside won't be converted into `Predicate<Entity>`.

~~Maybe turn the `filter` method into a vararg?~~

Or not - `filterSelector` uses `filter` method using a list of predicates.

MaxNeedsSnacks (approved):

KonSola5:
In addition:
- `EntityArrayList#filter` is inaccessible from JS by calling it plainly due to order of method lookup - Rhino first looks up "custom methods" (JS methods for Java lists), THEN the methods on the class. `filter` is one of them, so JS filter is used.
I think for now this should be renamed to something else, like `filterList` and remove buggy `oneFilter` (it breaks after finding one entity). Unless that's what the function was meant to do, IDK, there's no javadoc.
- Also, should the vanilla `hurt` (remapped to `damage`) be hidden and replaced by another method that inverts the parameters? This way, the effectively optional parameter (damage source) will be at the end, as it is usually in JS.

KonSola5:
New stuff:
- Remapped `Entity#setYBodyRot` and `Entity#getVisualRotationYInDegrees` to `setBodyYaw` and `getBodyYaw` respectively.
- Added `EntityKJS#kjs$isClientPlayer` and `EntityKJS#kjs$isServerPlayer` to check for a `LocalPlayer` or `ServerPlayer`, server player checks are quite frequent in scripts.
- Changed `ServerPlayerKJS#kjs$ban`'s third parameter to `Duration` (from `long`).
  - KubeJS has a type wrapper for `Duration`, so scripters can now also use string formats like `'2d'` to provide ban durations.
  - Not every `long` can be described using a JS number (double).

PR #1099, opened 2026-02-23
Fix custom mod names with EMI

JL2210:
### Description 

Previously EMI would just show the modid string if it couldn't find the fancy name. Now it works similarly to JEI and REI.

#### Example Script 

```js
Platform.mods.setModName("fff", "Flurben's Fantastic Foods")

StartupEvents.registry("item", event => {
  event.create("fff:cassava_seeds")
})
```

#### Other details 

Screenshot of fixed:
image


MaxNeedsSnacks:
Thank you for the PR! Since this is a bug fix, I will still be merging it to the 1.20 release, though I'm not entirely sure if I can cleanly port this code to 1.21.1 as well. If not, do you mind making an equivalent PR for 1.21 as well?

JL2210:
It doesn't look like these mixins or EMI changed any from 1.20 to 1.21. The only difference I see is common folders being merged since Fabric support was removed, but that would be a problem for everything

Issue #1105, opened 2026-03-19
[Bug] Incompatibility with Connector causes EvaluatorException | Error in 'ItemEvents.modification': Cannot convert ArrowFunction (1) => {...} to net.minecraft.world.food.FoodProperties


DeliciousBread481:
### Minecraft Version

1.20.1

### KubeJS Version

2001.6.5-build.16

### Rhino Version

2001.2.3-build.10

### Architectury Version

9.2.14

### Forge/Fabric Version

Forge 47.4.16

### Describe your issue

When I install `Connector`,the item.js will loaded with error
`[1932026 12:38:54.595] [Worker-Main-14/ERROR] [KubeJS Startup/]: …rhino.EvaluatorException: Cannot convert ArrowFunction (1) => {...} to net.minecraft.world.food.FoodProperties (startup_scripts:azicai/item.js#15)`

If I disabled the `Connector`,it doesnt happen.
So I check the source code,and found that:
Under normal circumstances, the Item.foodProperties field is private. KubeJS exposes two setter methods through Mixin and the @RemapPrefixForJS("kjs$") annotation:
1.kjs$setFoodProperties(FoodProperties properties) — directly sets a FoodProperties object
2.kjs$setFoodProperties(Consumer<FoodBuilder> consumer) — accepts a callback function that builds FoodProperties via FoodBuilder 
In JS, @RemapPrefixForJS maps kjs$setFoodProperties to setFoodProperties, which in turn generates the bean property foodProperties. So when you write item.foodProperties = food => { ... }, Rhino calls the setter method, matches the Consumer<FoodBuilder> overload, and the arrow function is correctly converted to a Consumer.
When the Connector mod is present, it widens the foodProperties field in the Item class from private to public. When Rhino's engine resolves the property, it finds a public field foodProperties (of type FoodProperties) and prefers direct field assignment over the setter method. At this point, Rhino attempts to directly convert the arrow function to the FoodProperties type, but since there is no type wrapper registered for it, the error is thrown

And If I change the:
ItemEvents.modification(event => {
&nbsp;&nbsp;event.modify('minecraft:rabbit_foot', item => {
&nbsp;&nbsp;&nbsp;&nbsp;item.foodProperties = food => {
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;food.hunger(2)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;food.saturation(0.1)
[...]

PR #1106, opened 2026-03-20
fix: register FoodProperties type wrapper to fix Connector mod compatibility 

DeliciousBread481:
Fix https://github.com/KubeJS-Mods/KubeJS/issues/1105

### Description 

Register a type wrapper for FoodProperties in BuiltinKubeJSPlugin.registerTypeWrappers(). When the incoming value is a JS function (BaseFunction), use FoodBuilder to construct the FoodProperties; when the incoming value is already a FoodProperties, return it directly.

This way, even if Connector widens the field to public and Rhino takes the direct field assignment path, the type wrapper will correctly convert the arrow function.

#### Example Script 

```js
ItemEvents.modification(event => {
    event.modify('minecraft:rabbit_foot', item => {
      item.foodProperties = food => {
          food.hunger(2)
          food.saturation(0.1)
          food.fastToEat()
      }
    })
  })
```

#### Other details 
The error cause I have said in the issue
Hope you can merged my PR,thanks⭐


MaxNeedsSnacks (approved):

MaxNeedsSnacks:
Gonna have to change this because it changed behaviour in existing scripts unintentionally; to set a raw food properties object, you can use `kjs$overrideFood()` now, while the setter for food properties still takes in a `Consumer<FoodBuilder>`

PR #1110, opened 2026-04-10
Fix `DataComponentWrapper` incorrectly handling component removal

andriihorpenko:
### Description
`DataComponentWrapper` incorrectly handles "!" prefix inside a data component notation, causing item string parsing to fail.
```java
Failed to read item stack from music_disc_cat[!jukebox_playable]: Unknown item component 'minecraft:' at position 15: ..._disc_cat[<--[HERE]
```

This happens because instead of removing "!" prefix before proceeding to component type parsing, `skipWhitespace` is called, leaving "!" prefix untouched. This PR simply swaps incorrect call with a `skip` function.

#### Example Script

```js
Item.of('music_disc_cat[!jukebox_playable]')
```


MaxNeedsSnacks:
Thanks for the contribution!

PR #1120, opened 2026-05-03
Small fix to allow for item quantity changes in existing recipes

lexxieblack:
Makes changing input/output item quantities possible.
This test script wouldn't have changed any of the output quantities prior to this:
```js
ServerEvents.recipes(e => {
    e.replaceOutput({ id: 'minecraft:furnace' }, 'minecraft:furnace', '2x minecraft:furnace')
    e.replaceOutput({ id: 'minecraft:crafting_table' }, 'minecraft:crafting_table', Item.of('minecraft:crafting_table', 3))
    e.replaceOutput({ id: 'minecraft:redstone_block' }, 'minecraft:redstone_block', { item: 'minecraft:redstone_block', count: 4 })

    e.replaceOutput({ id: 'minecraft:oak_planks' }, 'minecraft:oak_planks', '9x minecraft:oak_planks')
    e.replaceOutput({ id: 'minecraft:spruce_planks' }, 'minecraft:spruce_planks', '3x minecraft:diamond')
})
```
But now it does:
image
image

PR #1121, opened 2026-05-03
Small fix to allow for item quantity changes in existing recipes

lexxieblack:
Exact same changes as #1120


MaxNeedsSnacks:
See #1120

PR #1122, opened 2026-05-04
Small fix to allow for item quantity changes in existing recipes

lexxieblack:
Same outcome as #1120, but with a different implementation, as the code is much different in 1.20


MaxNeedsSnacks:
See #1120

PR #1131, opened 2026-05-25
Reworks `ItemBuilder` and `ItemModificationKubeEvent` for integration with IItemExtension and modification of items

Prunoideae:
### Description 

Separated the callbacks from `ItemBuilder` as a standalone class `ItemBehavior`, which allows the `ItemModificationKubeEvent` to modify the item's behavior (not registered by KubeJS) via the `ItemMixin`.

Also, since KubeJS has only supported Neoforge for quite some time, many methods from `IItemExtension` were patched by mixin to further extend the modifications of item behavior.

#### Example Script 

Modifies the diamond boot to make players walk on snow while wearing it.
```js
// startup_script
ItemEvents.modification(event => {
    event.modify('minecraft:diamond_boots', item => {
        item.canWalkOnPowderedSnow(true)
    })
})
```

image

#### Other details 

- The `IItemExtensionMixin` has a priority of `999`, so it applies earlier than defaults, while `1001` applies later than defaults. There are a few mods that share the same inject point, so using `1001` will prevent the mixin from working.

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/core/ItemKJS.java
+++ b/src/main/java/dev/latvian/mods/kubejs/core/ItemKJS.java
@@ -1,6 +1,6 @@
 package dev.latvian.mods.kubejs.core;
 
-import dev.latvian.mods.kubejs.item.ItemBuilder;
+import dev.latvian.mods.kubejs.item.ItemBehavior;
 import dev.latvian.mods.kubejs.item.ItemStackKey;
 import dev.latvian.mods.rhino.util.HideFromJS;
 import dev.latvian.mods.rhino.util.RemapForJS;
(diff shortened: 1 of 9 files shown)


MaxNeedsSnacks:
Thanks for the contribution! Do you think this is worth backporting to 1.20 or should we keep it for 1.21+ only?

Prunoideae:
I think the ItemBuilder -> ItemBehavior part is worth it if many people are staying at 1.20.1, but the IItemExtension part is not very easy, considering KJS 1.20.1 is still depending on arch.

MaxNeedsSnacks:
Sounds about right, do you mind backporting just that then? I am also thinking of backporting the ...Functions split I did in 26.1 to 21.1 if I haven't done that already

Prunoideae:
Ok, I will make another PR when it's done.

PR #1136, opened 2026-06-07
Update Neo to first stable version

wolfieboy09:
NeoForge has announced that NF version `26.1.2.71` is the first stable version, so update to that


wolfieboy09:
Seems like Neo introduced breaking changes, will fix that in a new pr

PR #1147, opened 2026-06-13
Fix ConsoleJS and ScriptType initialization cycle

liopyu:
This fixes a circular initialization bug between `ConsoleJS` and `ScriptType` before script consoles are ready. Minimal reproduction of this bug would be registering a KubeJS event group with `EventGroup.server()`, which initializes `ScriptType` while `ConsoleJS` is still constructing its instances, causing a startup crash.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 # changelog
 
 ## Unreleased
+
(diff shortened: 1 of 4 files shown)


MaxNeedsSnacks:
You know, all this really does for me is confirm that `ConsoleJS.X` shouldn't exist at all, so we'll just be removing them in KJS 9

PR #1152, opened 2026-06-16
Better fix for Client binding timing bug via Lazy init

liopyu:
This reverts #1145 in favor of a more robust lazy initialization of the Client binding, initializing a lazy reference and storing it on the first call it isn't null. The issue with the previous fix was that it added the binding later from client setup, but that could target a different Rhino context/scope than the one startup scripts were loaded with, so existing startup event callbacks still couldn't resolve `Client`. 

This keeps `Client` registered in the original script scope while avoiding the early `Minecraft.getInstance()` null timing issue.

### DIFF
--- a/src/main/java/dev/latvian/mods/kubejs/script/BindingRegistry.java
+++ b/src/main/java/dev/latvian/mods/kubejs/script/BindingRegistry.java
@@ -1,6 +1,11 @@
 package dev.latvian.mods.kubejs.script;
 
+import dev.latvian.mods.rhino.BaseFunction;
+import dev.latvian.mods.rhino.Context;
 import dev.latvian.mods.rhino.Scriptable;
+import dev.latvian.mods.rhino.ScriptableObject;
+
+import java.util.function.Supplier;
 
 public record BindingRegistry(KubeJSContext context, Scriptable scope) {
 	public ScriptType type() {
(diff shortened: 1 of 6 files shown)


src/main/java/dev/latvian/mods/kubejs/script/BindingRegistry.java
```diff
@@ -12,4 +17,32 @@ public void add(String name, Object value) {
 			context.addToScope(scope, name, value);
 		}
 	}
+
+	public void addLazy(String name, Supplier<?> supplier) {
```
MaxNeedsSnacks, line 26: this should use our own lazy type, which supports invalidation and doesn't need additional caching

src/main/java/dev/latvian/mods/kubejs/script/BindingRegistry.java
```diff
@@ -12,4 +17,32 @@ public void add(String name, Object value) {
...
+			}
+		});
+		descriptor.put(context, "enumerable", descriptor, true);
+		descriptor.put(context, "configurable", descriptor, true);
+
+		if (scope instanceof ScriptableObject object && descriptor instanceof ScriptableObject descriptorObject) {
+			object.defineOwnProperty(context, name, descriptorObject);
+		}
+	}
 }
```
MaxNeedsSnacks: this method should be documented (all it does is create a property with a custom getter, but still)

liopyu: Added Markdown comments as well as some useful inlined ones

src/main/java/dev/latvian/mods/kubejs/script/BindingRegistry.java
liopyu: Implemented KubeJS Lazy as well as let normal add() take over once the value is non-null, re-assigning it to a normal global after our lazy variable is no longer needed.

PR #1153, opened 2026-06-16
Add headless game-test/unit-test harness with script-side AssertJ assertions and coverage

hspragg-godaddy:
### Description

Grows the initial "one `BlockEvents.broken` test" into a full **headless test harness** for KubeJS — game tests, script-side AssertJ assertions, a bootstrapped JUnit layer, and merged coverage reporting — plus a small 26.1 port bug fix the harness surfaced.

**Game-test harness**

- `KubeJSGameTests` (the `testmod` entry point) stands up NeoForge's TestFramework; AlmostGradle's `testMod`/`gameTests`/`testFramework`/`jUnit` sets are enabled. `runGametest` boots a headless dedicated server that loads the `testmod` mod plus the KubeJS scripts under `src/test/kubejs/`.
- `TestRuntime` (bound into scripts by `TestRuntimePlugin`) bridges a script and the game test driving it: `pass`/`passed`/`count`/`clear`/`passStartup` markers, plus `check`/`verify` and `assertThat`/`assertDefined` (below).
- Event coverage across groups — each test drives a world action with a mock server player and asserts the KubeJS event fired (or that its event object resolves): `BlockEventTests`, `EntityEventTests`, `LevelEventTests`, `ServerEventTests`, `ItemEventTests`, `PlayerEventTests`. Events that can't fire on a headless dedicated server (client/GUI/networking, stage mutation) are deliberately omitted.

**Script-side AssertJ assertions**

- `TestRuntime.assertThat(...)` exposes typed AssertJ entry points to scripts (String/number/boolean/Object/Iterable), plus custom `KubeEntityEventAssert` (`hasEntityType`/`hasPlayer`/`hasNoPlayer`) and `LevelBlockAssert` (`hasId`/`hasProperty`) for event objects.
- Since the event dispatcher swallows exceptions thrown in handlers, assertions run inside `TestRuntime.check(id, () => …)`, which captures the failure; the paired game test calls `TestRuntime.verify(id)` to re-throw it on the game-test thread with AssertJ's message — so a wrong value fails the test instead of silently timing out.
- `kjs_reads.js` + `KjsReadTests` exercise the script-exposed `*KJS` convenience getters (EntityKJS/LivingEntityKJS/ServerPlayerKJS/LevelBlock/ItemStackKJS) and assert each resolves (`assertDefined`).
- `builder_fixtures.js` registers a spread of blocks/items/fluids to cover the builder classes during startup-registry load.

**Unit tests**

- `BootstrapExtension` runs `Bootstrap.bootStrap()` once so JUnit tests can hit registry-backed script API (wrappers, block lookups) without standing up a server.
- Over two dozen unit-test classes over pure/registry-backed helpers: `util` (ID, Tristate, TimeJS, TickDuration, IntBounds, CountingMap, Object2LongEntry, ErrorStack, TinyMap, JsonUtils, UtilsJS, NBTUtils, ListJS, RegExpKJS), wrappers (Item/Block/Color/NBT/Misc/Text), recipe components (Number/Boolean/String/Character), and MapColorHelper.

**Coverage**

- `jacoco` (0.8.15 — the first version supporting Java 25 class files) attached to both the JUnit `test` JVM and the `runGametest` server, scoped to `dev.latvian.mods.kubejs.*`. `coverageReport` merges both into one HTML+CSV report; `coverageSummaryJson` derives `coverage-summary.json` (instruction/branch/line %). A `Coverage` workflow runs both suites on PRs and posts the summary. Whole-mod instruction coverage is ~26% and rises with each added test.

**Bug fix**

- `DetectorBlock` failed to register on 26.1 (its block id was never stamped on `Properties`), which crashed world load; fixed, which also unblocks the detector game tests.

### Example Script

Server scripts can now assert on event objects, with failures surfaced to the driving game test:

```js
BlockEvents.broken(event => {
	TestRuntime.check('block.broken.assert', () => {
		TestRuntime.assertThat(event.block).hasId('minecraft:dirt');
		TestRuntime.assertThat(event).hasPlayer();
	});
});
```

The paired `@GameTest` breaks a dirt block, waits for the marker, then calls `TestRuntime.verify("block.broken.assert")` to re-throw any captured AssertJ failure on the test thread.

### Example Command

```bash
./gradlew runGametest          # headless game-test server
./gradlew test                 # JUnit unit tests
./gradlew coverageSummaryJson  # both suites + merged coverage report/JSON
```

Output lands in `build/reports/coverage/` (HTML, CSV, `coverage-summary.json`); game-test logs under `build/tmp/gametestRuns/gametest/logs/`; JUnit results under `build/reports/tests/test/`.

### Other details

- Adds a test-only `org.assertj:assertj-core` dependency (`testImplementation`).
- The harness is server-side only; client/GUI/rendering/recipe-viewer bindings aren't exercised.
- Most coverage comes from the mod booting during the game test, so the number grows as more targeted tests are added.
- The PR-comment coverage step only runs for branches in the main repository, since fork PRs receive a read-only token.


build.gradle.kts
```diff
@@ -23,7 +27,11 @@ almostgradle.setup {
 	withAccessTransformerValidation = !runningInCI
 
 	tests {
-
+		enabled = true
```
rlnt: This can be omitted. It's implicitly set in AlmostGradle when the `tests {}` block is used. It's used for internal state handling.

build.gradle.kts
```diff
@@ -61,6 +69,15 @@ neoForge {
 		from(file("interfaces.json"))
 		publish(file("interfaces.json"))
 	}
+
+	runs {
+		// AlmostGradle's gametest run launches as a plain server here; use Neo's headless type and run only our tests.
```
rlnt: Is this required because you want to run the unit tests when the game tests are running? I could add a flag to AG that enables this automatically if required.

src/test/java/dev/latvian/mods/kubejs/testmod/block/BlockEventTests.java
```diff
@@ -0,0 +1,26 @@
...
+import net.neoforged.testframework.annotation.ForEachTest;
+import net.neoforged.testframework.annotation.TestHolder;
+import net.neoforged.testframework.gametest.EmptyTemplate;
+import net.neoforged.testframework.gametest.GameTest;
+
+@ForEachTest(groups = "kubejs.block.event")
+public class BlockEventTests {
+	@GameTest
+	@EmptyTemplate(floor = true)
+	@TestHolder(value = "block_broken_dirt", description = "KubeJS BlockEvents.broken fires when a player breaks dirt")
```
rlnt, line 37: Why do you use the framework integration here? A standard game test would be enough. Usually, the framework is used to wire in Neoforge events. The `EmptyTemplate` annotation can be used on normal game tests as well.

If this is a style choice, so that all tests use the same format, that's fine. I just wanted to know if there is a specific technical reason.

hspragg-godaddy: There wasn't any specific technical reason.
When I was using NeoForge source as a reference, I really liked the pipeline approach and switched to the GameTest format.
[Earlier in the commit chain](https://github.com/KubeJS-Mods/KubeJS/pull/1153/commits/4cd44709a7a0d07fbaf19dce5cae3332f73b8b39) I was using the alternative format and felt it didn't look as nice as the alternative.

build.gradle.kts
hspragg-godaddy: Without this section it starts up a regular minecraft server rather than a gametest server. Even kicks off the old server GUI.

build.gradle.kts
hspragg-godaddy: Removed in latest commit

build.gradle.kts
rlnt: I have now fixed this in Almost Gradle 2.2.0. When you update, this should no longer be necessary. The game test property name was updated, which I must have missed. Please report it the next time you discover something like this. 😄

I also added a new option to run the unit tests in the game test run config, which you configured here as well.
https://github.com/AlmostReliable/almostgradle#unit-tests-in-game-tests

However, as written in the README, I would not encourage enabling this behavior. Unit tests are invoked in the `test` Gradle task, which is part of the `check` Gradle task. And `check` is invoked when `build` is run. So a successful build already guarantees that all unit tests passed.

Hunter19823: I just pushed changes for the latest version and removed the section like you had suggested.
I also isolated the game tests to not run on build in favor of just the tests running on build.
That way builds don't trigger game tests.

src/test/java/dev/latvian/mods/kubejs/unittest/BlockWrapperTest.java
```diff
@@ -0,0 +1,42 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+@ExtendWith(BootstrapExtension.class)
+public class BlockWrapperTest {
+	@Test
+	void resolvesVanillaBlock() {
+		assertThat(BlockWrapper.getBlock(Identifier.parse("minecraft:stone"))).isSameAs(Blocks.STONE);
+		assertThat(BlockWrapper.getId(Blocks.STONE)).isEqualTo(Identifier.parse("minecraft:stone"));
+	}
+
```
MaxNeedsSnacks: (This applies to all wrapper tests)

In my eyes, all of these tests should use a JS script with either a custom event that is fired by the harness or just a top-level immediately invoked function expression, checking for equality and other assertions can either be done in the JS side or back on the Java side by returning from JS

src/test/java/dev/latvian/mods/kubejs/unittest/BlockWrapperTest.java
```diff
@@ -0,0 +1,42 @@
...
+	void blockStateStringFormats() {
+		assertThat(BlockWrapper.toBlockStateString("minecraft:stone", Map.of())).isEqualTo("minecraft:stone");
+		assertThat(BlockWrapper.toBlockStateString("minecraft:furnace", Map.of("lit", "true")))
+			.contains("minecraft:furnace").contains("lit").contains("true");
+	}
+
+	@Test
+	void idPredicateBuilds() {
+		assertThat(BlockWrapper.id(Identifier.parse("minecraft:stone"))).isNotNull();
+	}
```
MaxNeedsSnacks: related to the above, this test is frankly kind of redundant without any actual type wrapping happening (what are we testing here? that a Java constructor... works)

src/test/java/dev/latvian/mods/kubejs/unittest/CharacterComponentTest.java
```diff
@@ -0,0 +1,38 @@
...
+	@Test
+	void displayStringWrapsInSingleQuotes() {
+		assertThat(CharacterComponent.CHARACTER.toString(null, 'a')).isEqualTo("'a'");
+	}
+
+	@Test
+	void metadataIsStable() {
+		assertThat(CharacterComponent.CHARACTER.typeInfo()).isEqualTo(TypeInfo.CHARACTER);
+		assertThat(CharacterComponent.CHARACTER.toString()).contains("character");
+	}
```
MaxNeedsSnacks: These tests seem similarly redundant because the code they check is just so trivial...

src/test/java/dev/latvian/mods/kubejs/unittest/ColorWrapperTest.java
```diff
@@ -0,0 +1,37 @@
+package dev.latvian.mods.kubejs.unittest;
+
+import dev.latvian.mods.kubejs.plugin.builtin.wrapper.ColorWrapper;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ColorWrapperTest {
```
MaxNeedsSnacks: see above regarding wrapper tests (this is the only time I'll refer back to that comment)

src/test/kubejs/server_scripts/kjs_reads.js
MaxNeedsSnacks, line 1: This doesn't seem like a useful test; instead, we should have scripts that listen to certain events, maybe do things with the entity based on its type (say you spawn a cow and a zombie, the zombie should be given a held item which can be checked in the Java side, the cow should be moved south by 5 blocks etc.)

src/test/kubejs/startup_scripts/builder_fixtures.js
```diff
@@ -0,0 +1,128 @@
...
+		.steppedOn(cb => {})
+		.fallenOn(cb => {})
+		.afterFallenOn(cb => {})
+		.exploded(cb => {})
+		.rotateState(cb => {})
+		.mirrorState(cb => {})
+		.rightClick(cb => {})
+		.defaultState(cb => {})
+		.placementState(cb => {})
+		.canBeReplaced(cb => true);
```
MaxNeedsSnacks: Are these tested?

src/test/kubejs/startup_scripts/builder_fixtures.js
```diff
@@ -0,0 +1,128 @@
...
+		.maxStackSize(16)
+		.glow(true)
+		.tooltip('Coverage item')
+		.fireResistant()
+		.containerItem('minecraft:bucket')
+		.burnTime(200)
+		.food(4, 0.3)
+		.disableRepair()
+		.rarity('rare')
+		.tag('minecraft:planks');
```
MaxNeedsSnacks: see above, are these tested? I hesitate to call these integration tests if the methods are never really integrated into "real" game tests

MaxNeedsSnacks (changes requested):

src/test/kubejs/server_scripts/syntax_checks.js
MaxNeedsSnacks, line 1: these tests are unnecessary imo and should be part of rhino itself if anything

src/test/kubejs/server_scripts/wrapper_checks.js
```diff
@@ -0,0 +1,87 @@
...
+// boundary (TestRuntime.as*), forcing the registered type wrapper to run, then asserts the coerced
+// object. These run once on the first server tick - not at script load - so registry-backed
+// conversions (item components) are bound; WrapperTests waits for and verifies the captured results.
+
+let wrapperChecksRun = false;
+
+ServerEvents.tick(event => {
+	if (wrapperChecksRun) {
+		return;
+	}
```
MaxNeedsSnacks, line 11: why use a single-run ServerEvents.tick instead of firing a custom event when the game test world is ready?

hspragg-godaddy: Exceptions were being thrown on first tick, this was a mitigation to that.

PR #1168, opened 2026-07-29
Update Plugin Registration to a JSON based system

wolfieboy09:
### Description
This PR supersedes #1150 by having a JSON-based plugin registration system. 
By having a JSON-based system for plugins, having everything in one file should be easier to maintain for everyone

This rewrite does keep legacy support for the text-based system. Class filter txt is still supported, as big filters should be in there instead.

#### Other details
For moving the entire class filter to the plugin JSON, this Python script can be ran in the root folder:
<details>
<summary>Python migration script</summary>

```py
import json

TXT_PATH = "src/main/resources/kubejs.classfilter.txt"
JSON_PATH = "src/main/resources/kube.plugin.json"

allow: list[str] = []
deny: list[str] = []

# Creates a list of allowed and denied classes.
# Removes the space after the '+' and '-'
with open(TXT_PATH, "r") as file:
    content = file.read()

for line in content.splitlines():
    lne = line.strip()
    if not lne or lne.startswith("#"):
        continue
    if lne.startswith("+"):
        allow.append(lne[1:].strip())
    elif lne.startswith("-"):
        deny.append(lne[1:].strip())

with open(JSON_PATH, "r") as file:
    data = json.load(file)

plugin = data["plugins"][0]
class_filter = plugin.setdefault("class_filter", {})

if allow:
    class_filter["allow"] = allow
if deny:
    class_filter["deny"] = deny

with open(JSON_PATH, "w") as file:
    json.dump(data, file, indent=2)
    file.write("\n")

print("Class filter migration completed. You can now delete the class filter txt file.")
```

</details>
If this PR does get merged, I can backport this to `1.21.1` if needed.

#### Checklist
- [x] All fields work as intended
- [x] Update README.md with all the fields
- [x] Legacy text-based plugins still load


Notenoughmail:
java datagen for this would be *very* nice :), would more easily guarantee plugins are `KubeJSPlugin`s

also, I'd personally use a codec for `PluginFileData`[^1] and consequentially `PluginEntry` and `ClassFilterData`, but I will leave that to Max/Lat to decide

[^1]: imagine, if you will, `Codec<Class<? extends KubeJSPlugin>>`

wolfieboy09:
I have added codecs for `PluginFileData`, `PluginEntry`, and `ClassFilterData` to make everything easier. I was not sure how to have `Codec<Class<? extends KubeJSPlugin>>`, so I just have it as a `Codec<PluginFileData>`

Notenoughmail:
```java
Codec.STRING.comapFlatMap(str -> {
  try {
    var clazz = Class.forName(str);
    return DataResult.success(clazz.asSubclass(KubeJSPlugin.class));
  } catch (ClassNotFoundException | ClassCastException e) {
    return DataResult.error(() -> "Error decoding plugin class: " + e.getMessage());
  }
}, Class::getName)
```
should do it

wolfieboy09:
May I ask what the use case would be for having access to the `Class<? extends KubeJSPlugin`? I have it loading plugins just fine as before, but changing it to store the class.

wolfieboy09:
The snippet you provided did work, just had to modify it a little to prevent everything else from not loading

Notenoughmail:
> May I ask what the use case would be for having access to the `Class<? extends KubeJSPlugin`? I have it loading plugins just fine as before, but changing it to store the class.

It's less having access to it and more being ideomatic about codec usage *and* having the proper type limit when making the plugin data in datagen scenarios

wolfieboy09:
Ohh okay, I see why that's a lot more proper to have

wolfieboy09:
NOO HOW DID I MANAGE TO DO THIS :sob:

PR #1169, opened 2026-08-03
move HolderSet wrapping to TypeWrappers

LLytho:
fixes HolderSet wrapping when regex is being used after <https://github.com/kube-mods/rhino/commit/d46e972a2a4e9d26b9969c9cccfd9eac078884a5>


MaxNeedsSnacks:
nice, thank you; can you create a version for 1.21 as well, or rebase this onto that?

src/main/java/dev/latvian/mods/kubejs/script/KubeJSContext.java
Notenoughmail, line 110: Should `HolderSer` be removed from here?

MaxNeedsSnacks: actually, yeah it probably should; we probably don't ever even reach this path anymore thanks to the type wrapper

MaxNeedsSnacks:
thanks for the fix, superseded by 1d2fce2

